PR #59, opened 2015-01-14
Pass THCState into THC functions.

dominikgrewe:
Changes needed for torch/cutorch#99


soumith:
this PR actually does not load cunn, because utils.c is never actually compiled into libcunn.so
Fixed via #62

PR #60, opened 2015-01-16
Added SpatialAdaptiveMaxPooling

fmassa:
https://github.com/torch/nn/issues/141

### DIFF
--- a/SpatialAdaptiveMaxPooling.cu
+++ b/SpatialAdaptiveMaxPooling.cu
@@ -1,0 +1,374 @@
+
+#define CUDA_MAX_THREADS 1024   // this is safe, in reality 256 is our limit
+
+/*
+ * Description:
+ *    this function adaptively maxpools an input 4D tensor along dimensions 2 and 3
+ *    4D input, 4D output, 4D argmax x and y 
+ */
+__global__ void adaptivemaxpool(float *input, float *output, float *indices_x, float *indices_y,
+                        int input_n, int input_h, int input_w,
+                        int output_h, int output_w)
+{
+  // iterators
+  int xx, yy;
+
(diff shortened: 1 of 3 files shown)


SpatialAdaptiveMaxPooling.cu
```diff
@@ -0,0 +1,372 @@
+
+#define CUDA_MAX_THREADS 1024   // this is safe, in reality 256 is our limit
```
soumith: can you undef this at the end of the file.

soumith:
just one small comment, otherwise looks good i think

SpatialAdaptiveMaxPooling.cu
szagoruyko: I guess ```const int CUDA_NUM_THREADS = 1024;``` is better

soumith: hmm, if it is scoped in a function, yes. not in global scope.

szagoruyko: no it's used inside in the kernel, but it's like this in SpatialMaxPooling.cu, SpatialAveragePooling.cu and SubSampling.cu, without undef

soumith:
torch/nn#141 had a build-break. can you please retest that stuff and send it in. i'll test it on my end and then merge (this PR as well)

szagoruyko:
ah we forgot to commit the main file SpatialMaxPooling.lua, will fix it now sorry

szagoruyko:
checked nn version, looks good, checking cunn

szagoruyko:
ok both nn and cunn versions work, tested on 3 machines with unit tests, ready to be merged.

soumith:
thanks fmassa and szagoruyko!

PR #61, opened 2015-01-16
undef CUDA_MAX_THREADS in MaxPooling, SubSampling and AveragePooling

szagoruyko:
small fix as discussed here https://github.com/torch/cunn/pull/60


soumith:
thanks!

PR #66, opened 2015-02-18
Add LogSigmoid module.

dominikgrewe:
Add matching module for nn.LogSigmoid and a test to verify it behaves
the same.


dominikgrewe:
This was a patch I had for a while. I just realized that I didn't rebase this after the THCState change, so it won't compile. Really sorry!! Fix is on its way.

PR #71, opened 2015-03-06
Fix SpatialAveragePooling.

dominikgrewe:
Use atomic version of subgradinput in the non-batch case if kH != dH or
kW != dW.

### DIFF
--- a/SpatialAveragePooling.cu
+++ b/SpatialAveragePooling.cu
@@ -262,8 +262,15 @@ static int cunn_SpatialAveragePooling_updateGradInput(lua_State *L)
     dim3 threads(32,8);
 
     // run updateGradInput kernel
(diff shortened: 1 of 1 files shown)


soumith:
hmm, i thought I made this change before. strange.

soumith:
strange indeed. i did the same in this commit https://github.com/dominikgrewe/cunn/commit/fb938461c0e4277e9116e572655fa6388d5f0a9c but it disappeared after even though none of the commits after diff it

dominikgrewe:
You only added it to the batch-case, but forgot about the non-batch case.

soumith:
ohh ok, thanks. i was losing it.

dominikgrewe:
No worries :)

PR #73, opened 2015-03-10
Added DataParallelTable.  Added docs.

jonathantompson:
See the docs for an explanation and usage.

After chatting with @soumith, it seems like he would prefer that the lua source goes in cunn, rather than nn (which makes sense to me since this is a GPU only related module).  It means that a) CMakeLists needs to change so that all lua source is copied and b) I needed to add docs.

### DIFF
--- a/DataParallelTable.lua
+++ b/DataParallelTable.lua
@@ -1,0 +1,506 @@
+local gpu_local_copy_buffers = {}
+local base_gpu_index = 1  -- A constant
+
+-- *****************************************************************************
+-- Helper Functions
+-- *****************************************************************************
+-- queryGPUDeviceId - Function to query a tensor or table for the 
+-- GPUID.  For tables we will search the table for CudaTensors, query their 
+-- device and make sure the deviceIds of ALL CudaTensors are on the same GPU.
+local function queryGPUDeviceId(object)
+  if torch.type(object) == 'torch.CudaTensor' then
+    return object:getDevice()
+  end
+
+  local deviceId
+
+  -- Try finding a parameter
+  local stack = {}  -- explicit stack to recurse on tables
+  for key, param in pairs(object) do
+    if key ~= 'modules' then
+      stack[#stack+1] = param  -- Push onto the stack
+    end
+  end
+  while #stack > 0 do
+    local param = stack[#stack]; stack[#stack] = nil  -- Pop the stack
+    if (torch.type(param) == 'table') then
+      for i = 1, #param do stack[#stack+1] = param[i] end  -- Push onto stack
+    elseif (torch.type(param) == 'torch.CudaTensor') then
+      if (torch.numel(param) > 0) then
+        -- Empty tensors are always on GPU "0"
+        local cur_id = param:getDevice()
+        if deviceId == nil then
+          deviceId = cur_id
+        else
+          assert(deviceId == cur_id, 
+            'Found CudaTensor instances from different devices')
+        end
+      end
+    end
+  end
+
+  return deviceId
+end
+
+-- Get an avaliable GPU buffer for asyncGPUCopy.  It is used when the GPU tensor
+-- is not contiguous.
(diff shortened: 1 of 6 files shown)


jonathantompson:
It would be nice if we could do a proper code-review on this one.  I've been using it in my production code for about a week now and it seems to be OK.  I've also run cuda-memcheck on test_DataParallelTable.lua, but I'm not 100% sure if there aren't other nasty corner cases.

soumith:
i have a curious idea  
i think net:syncParameters() can be avoided in a way:

keep a self.synced boolean variable  
in accGradParameters mark it to false  
in :updateOutput, if it is false, force-sync.  
what do you think?

jonathantompson:
Yeah, I thought that might work...  However, it assumes that you're calling
forward, backward, forward, etc.  One might be inclined to call backward
multiple times before performing the gradient step (I can't think of a
reason of the top of my head, but its a possibility).  So to keep it
general, I figured the syncParameters call seemed to be the easiest API.

On Tue, Mar 10, 2015 at 12:45 PM, Soumith Chintala <notifications@github.com

soumith:
even in your use-case there's a clear boundary on when to sync and when not to sync right, i.e. in the case of forward-backward-backward-backward-forward. 
I dont mind this new change, but it would become 100% natural to use DataParallelTable if something like that is introduced.

jonathantompson:
Any updates?  Seems silly to waste all that work...

soumith:
i've reviewed the code. just debating on how to make it seamless to the user to use. Is there a possibility that you can remove syncParameters() by integrating it into the forward/backward phases? If not it's ready for merge.

jonathantompson:
OK, so I was actually thinking about taking it one step further.  Speaking to Wojciech, it seems like the magic hack to get DataParallel scaling is to only update the weights every few mini-batches.  ie you let the models diverge and then every few batches you average the parameters.

I haven't tried it but I plan to do so this week. I would also need to train a model to completion to make sure the performance is not degraded... Therefore, I would like to keep the API for this stage the way it is and maybe make it even more "user-unfriendly", but enable the user to sync the parameters when they choose to do so.  What do you think?  You probably wont like it, but personally I'd rather open up the sync() function to the user so that a) they have the freedom to do as described and b) they are aware of the PCIe overhead of moving weights around (which from my experience so far is pretty bad).

I dream of having training be 2x faster :-)  Right now I'm getting 1.6x faster on 2 GPUs which is a little disappointing.

DataParallelTable.lua
```diff
@@ -0,0 +1,506 @@
...
+-- setDeviceSafe - Avoid redundant calls to setDevice
+local function setDevice(gpuid)
+  if (cutorch.getDevice() ~= gpuid) then
+    cutorch.setDevice(gpuid)
+  end
+end
+
+-- Asynchronous copy from source to dest from GPU to GPU.
+-- This is borrowed (with modifications) from fbcunn.
+local function asyncGPUCopy(dest, source)
```
dominikgrewe: Some (maybe all?) of this will be done by the new version of copy in torch/cutorch#120. So this could be simplified once that PR has been merged.

jonathantompson: Sounds good to me.

soumith:
@jonathantompson your comments sound good. I'll merge it at the current state then.
And wrt comments about longer times between sync, Sixin is the man for that. Go talk to him. http://arxiv.org/abs/1412.6651

DataParallelTable.lua
soumith: The new copy has landed. Jono can you update this PR and I'll merge it in? Sorry for the delay

soumith:
This has waited long enough, I've merged it in. Jono see my comment inline, the new copy kernels have landed.

jonathantompson:
Sorry, it's been a hectic weekend.  I'll try and get around to it when I
have time.

PR #74, opened 2015-03-13
Fix SpatialConvolutionCUDA:accGradParameters.

dominikgrewe:
It used to overwrite the gradients instead of accumulate them.

### DIFF
--- a/SpatialConvolutionCUDA.cu
+++ b/SpatialConvolutionCUDA.cu
@@ -177,7 +177,7 @@ static int cunn_SpatialConvolutionCUDA_accGradParameters(lua_State *L)
     nOutputPlane, outputHeight, outputWidth,
     kH, kW,
     -floor((double)padding/2), dW,
-    0, scale, partialSum
+    1, scale, partialSum
   );
 
   return 0;


dominikgrewe:
gradWeight doesn't change when calling backward multiple times:

```lua
require 'cutorch'
require 'cunn'

local conv = nn.SpatialConvolutionCUDA(3, 16, 4, 4, 1, 1):cuda()
local input = torch.CudaTensor(3, 16, 16, 4):uniform()
local gradOutput = torch.CudaTensor(16, 13, 13, 4)

--local conv = nn.SpatialConvolutionMM(3, 16, 4, 4, 1, 1):cuda()
--local input = torch.CudaTensor(4, 3, 16, 16):uniform()
--local gradOutput = torch.CudaTensor(4, 16, 13, 13)

conv:forward(input)
conv:backward(input, gradOutput)
print(conv.gradWeight:norm())

conv:backward(input, gradOutput)
print(conv.gradWeight:norm())

conv:backward(input, gradOutput)
print(conv.gradWeight:norm())
```

soumith:
at some point in the future this should be removed, because ccn2's version is superior and faster.

PR #75, opened 2015-03-19
Handle ClassNLLCriterion weights on gpu

mlosch:
Attention: Changes the interface of updateOutput and updateGradInput due to required weights.
Requires an update of ClassNLLCriterion.lua in the nn package. 
Otherwise fails at attempting to access a weight tensor.
(See PR github.com/torch/nn/pull/197 which updates ClassNLLCriterion.lua accordingly)


soumith:
Hey max, look at my comments on the PR in nn and let me know if you have any questions.

soumith:
closed via #120

PR #76, opened 2015-03-20
in-place threshold, sqrt stability fixes, faster logsoftmax

soumith:
This PR has to wait for https://github.com/torch/cutorch/pull/120 to merge, as the LogSoftMax depends on some code from that.

PR #77, opened 2015-04-03
Pointwise layers use apply kernels

szagoruyko:
Uses apply kernels instead of thrust, so all newContiguous() calls are removed.
It seems to me that there could be written a general Pointwise module (as in  cudnn) to which only forward and backward functors would be given, instead of copy-pasting.
Compile time increased.
Includes Sqrt.cu fix from https://github.com/torch/cunn/pull/76

### DIFF
--- a/Abs.cu
+++ b/Abs.cu
@@ -1,10 +1,11 @@
 #include "utils.h"
+#include "THCApply.cuh"
 
 struct absupdateOutput_functor
 {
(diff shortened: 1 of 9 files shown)


soumith:
rebased and merged via https://github.com/torch/cunn/commit/f809fb65c40b923e654056894722096001d2a41b

this is great.

PR #79, opened 2015-04-07
multi-GPU checks, adheres to new cutorch Stream API, in-place threshold

soumith:
This makes all of cunn fully adhere to the streams API: https://github.com/torch/cutorch/pull/128

Also added multi-GPU checks in every single module of cunn.

Also gets in the PR https://github.com/torch/cunn/pull/76


soumith:
it's a very mechanical and long PR, already been through review internally, dont spend time reviewing it unless really curious.

PR #83, opened 2015-04-30
Add TransferDevice module.

adamlerer:
With cutorch auto-mode, multi-gpu models can be run without explicitly
updating the device. However, copies must be explicit, so a
TransferDevice module is necessary to make the copy.

Here's an example multi-GPU module (added as a test):

```lua
cutorch.setDevice(0)

net = nn.Sequential()
for i = 1,3 do
  net:add(nn.Linear(1000,1000):cuda(i))
  net:add(nn.SoftMax():cuda(i))
  net:add(nn.TransferDevice(i, i+1)) -- a new little shim in cunn
end

local input = torch.CudaTensorOn(1, 1000)
local output = net:forward(input)
print(output:getDevice())

local gradOutput = output/100
local gradInput = net:backward(input, gradOutput)
print(gradInput:getDevice())
```

### DIFF
--- a/TransferGPU.lua
+++ b/TransferGPU.lua
@@ -1,0 +1,36 @@
+local TransferGPU, parent = torch.class('nn.TransferGPU', 'nn.Module')
+
+function TransferGPU:__init(srcDev, dstDev)
+   parent.__init(self)
+
+   self.srcDev = srcDev
+   self.dstDev = dstDev
+
+   self.gradInput = torch.CudaTensorOn(srcDev)
+   self.output    = torch.CudaTensorOn(dstDev)
+
+end
+
+function TransferGPU:updateOutput(input)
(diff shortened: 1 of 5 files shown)


soumith:
if it's under the nn namespace, make it nn.TransferGPU, TransferDevice is unclear what device we talk about.

soumith:
also add doc to cunn readme (which exists! ha)

adamlerer:
@soumith done thanks!

soumith:
this mixes the TransferGPU stuff with cudaOn stuff. I'll wait on the cudaOn PR to be merged first.

adamlerer:
@soumith my latest commit puts all the `cudaOn()` functionality into cunn, so this can be merged. I will abandon the nn PR.

soumith:
lgtm! thanks.

PR #86, opened 2015-05-05
PReLU cuda implementation

adampolyak:
Added cuda support for PReLU jointly with @szagoruyko.
Requires nn pull request torch/nn#258 to work.

### DIFF
--- a/PReLU.cu
+++ b/PReLU.cu
@@ -1,0 +1,254 @@
+#include "THCApply.cuh"
+#include "THCReduce.cuh"
+#include "utils.h"
+
+// CUDA: grid stride looping
+#define CUDA_KERNEL_LOOP(i, n)                        \
+  for (int i = blockIdx.x * blockDim.x + threadIdx.x; \
+      i < (n);                                       \
+      i += blockDim.x * gridDim.x)
+
+struct PReLUUpdateOutput {
+  float* weight_;
+
+  PReLUUpdateOutput(float* weight): weight_(weight) {}
+
+  __device__ __forceinline__ void operator()(float* out, float* in) {
+    float x = *in;
+    *out = (x > 0) ? x : weight_[0] * x;
+  }
+};
+
+__global__ void preluForward(float *output, const float *input, const float *weight,
+    int n, int nOutputPlane, int dim)
+{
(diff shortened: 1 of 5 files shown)


szagoruyko:
it might be possible to remove one buffer if there was THCudaTensor_reduceDim which allows an additional functor to modify output.
another thing is backward, we made it reuse gradInput and them recompute it to restore the state, because otherwise it would require an additional buffer size of the input, or lot's of big kernels which do reduceDim but take two tensors on input.

soumith:
you need to rebase the prelu commits back to top of master. if you notice, they are mixed, and the current PR shows commits that are already merged into trunk as well.

adampolyak:
fixed

PReLU.cu
```diff
@@ -0,0 +1,263 @@
+#include "THCApply.cuh"
+#include "THCReduce.cuh"
+#include "utils.h"
+
+// This is bad, because the following is defined in SpatialConvolution.cu ...
```
soumith: can you #undef at the end of SpatialConvolution.cu , and uncomment these

szagoruyko: btw would it be better to separate the compilation of each file in cunn? take some time now

dominikgrewe: I agree we should compile each file separately. We do it in cutorch now, which enables us to compile files in parallel.

nicholas-leonard: How do you accomplish that? Curious.

soumith: @nicholas-leonard https://github.com/torch/demos/blob/master/train-on-housenumbers/train-on-housenumbers.lua#L243
$(MAKE) -j$(getconf _NPROCESSORS_ONLN) install

adampolyak:
Added undef where required. Thing is, we are still using CUDA_NUM_THREADS and GET_BLOCKS which are defined in SpatialConvolutionMM.cu. By the way, VolumetricConvolution is also using those.

soumith:
ok, i'll have to do a cleanup pass on the macro defs, and splitting the compilation.

soumith:
if you squash your commits now, its ready to merge.

adampolyak:
done

soumith:
Thanks Adam!

PR #90, opened 2015-05-20
Lookup table GPU implementation based on LookupTableGPU from fbcunn

colesbury:
This is a CUDA implementation of nn.LookupTable based on LookupTableGPU from fbcunn.

It doesn't support "scaleGradByFreq" option. If anyone wants that on the GPU, I can implement it but it'll be a bit of work.

### DIFF
--- a/LookupTable.cu
+++ b/LookupTable.cu
@@ -1,0 +1,235 @@
+#include "utils.h"
+#include <thrust/device_ptr.h>
+#include <thrust/execution_policy.h>
+
+#ifndef DIVUP
+#define DIVUP(x, y) (((x) + (y) - 1) / (y))
+#endif
+
+const int WARP_SIZE = 32;
+
+__device__ __forceinline__ bool warpHasCollision(int val) {
+  // Compare our value to the values stored in the next 16 lanes,
+  // wrapping around at 32. If any pair of values is the same than
+  // there is a collision in the warp.
+  bool dup = 0;
+  const int laneId = threadIdx.x % 32;
+
+#if __CUDA_ARCH__ >= 300
+  #pragma unroll
+  for (int i = 1; i <= 16; i++) {
+    dup |= (__shfl(val, (laneId + i) % 32) == val);
+  }
+#else
+  volatile __shared__ int values[128];
+  values[threadIdx.x] = val;
+  const int offset = threadIdx.x - laneId;
+
+  #pragma unroll
+  for (int i = 1; i <= 16; i++) {
+    dup |= (values[offset + ((laneId + i) % 32)] == val);
+  }
+#endif
+
+  return __any(dup) != 0;
+}
+
+__global__ void cunn_LookupTable_accGradParametersKernelByFeature(
+  float *input, float *gradOutput, float *gradWeight, float scale, long numel, long stride) {
+
+  const int featureDim = blockIdx.x * 4 + threadIdx.x / 32;
+  if (featureDim >= stride) {
+    return;
+  }
+
+  // The strategy here is that each warp handles a single feature
+  // dimension.
+  // Within that feature dimension, points in the [batch][element]
+  // dimension can overlap, and we need to determine if threads want
+  // to add to the gradient in a colliding manner.
+  // Typically one would use floating-point atomicAdd() to resolve
+  // these collisions, but that is non-deterministic if there are
+  // collisions. Non-determinism for this code is really bad,
+  // especially in RNNs, and is prone to snowballing error.
+  // In order to get a deterministic order of execution, we handle
+  // non-colliding updates separately from colliding ones. Colliding
+  // updates are serialized in their order of execution by using the
+  // warp-wide collision detector `warpHasCollision`.
+  const int laneId = threadIdx.x % 32;
+  for (int i = laneId; i < numel; i += WARP_SIZE) {
+    int weightIndex = (int) (input[i] - 1);
+    float update = gradOutput[i*stride + featureDim] * scale;
+
+    // Check for collision
+    if (warpHasCollision(weightIndex)) {
+      // Run all lanes sequentially; warp divergence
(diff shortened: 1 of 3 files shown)


nicholas-leonard:
Awesome.

AjayTalati:
Super Awesome :+1:

zhangxiangxiao:
Does this have to use sm_30? Lots of people are still running torch on GTX 580 (sm_20), especially for projects on NLP which is what this module might be mostly used for.

soumith:
it's using the shuffle instruction which is only available in sm30. There could be a slower sort for back-compatible devices. @zhangxiangxiao do you think you'd be interested in doing that?

zhangxiangxiao:
Thanks for the clarification @soumith! I would have loved to do it but I do not think I understand the code nor would I have time. But I think it would be pretty easy to call the old code in case sm_20 is used, yes? For example we can disable compiling this module if sm_20 is detected and use the old method in the lua call by detecting whether input.nn.LookupTable_updateOutput exists.

soumith:
@wickedfoo @colesbury any simple fix for backcompat to sm_20? or even a much slower kernel....

wickedfoo:
The solution taken is in order to make the accGrad execute deterministically (avoiding atomicAdd).

A version based on sparse matrix-vector techniques exists as well, which is way faster for larger problem sizes but much slower for typical problem sizes in use.

Let's do this internally @soumith @colesbury and we'll offer that up.

colesbury:
I've made a few changes. This now compiles with sm_20 and sm_30.

I've included two implementations of accGradParameters. One is much faster for small inputs and the other is much faster for large inputs. We have some researchers which use a large (~20K) input size.

The second implementation requires sorting the inputs. The Modern GPU sorting implementation is much faster for inputs of size 1K-100K than the thrust::sort functions.

I've also added scaleGradByFreq option, which was requested by David.

wickedfoo:
@colesbury we had a little discussion today about this but I don't know if I'd go so far as requiring ModernGPU, and if so, maybe it should be as a separate dependency that people pull in from its source? Otherwise, it's just going to rot. If it is a separate dependency, that's annoying as well, since it isn't well packaged and seems to be somewhat orphaned by Nvidia.

How much slower than Thrust is it? Try using THCudaTensor_sort() instead, and if the Thrust sort is slow (it uses my small bitonic sort for small sizes, Thrust for larger sizes), let's address it in cutorch instead of this one-off. Requiring both CUB and ModernGPU in cutorch would be useful (there are lots of places where it would have been useful), but I think we should fix/implement a better sort in the core library rather than here.

cc @soumith

colesbury:
Here are some sorting times and the time for forward + backprop with LookupTable:

Input Size | ModernGPU | Thrust | THCudaTensor_sort [1] | LookupTable Forward+Backprop [2]
------------- | ----------------------- | --------------- | --------------------------- | -------------
32 | 41 μs | 680 μs | 55 μs | 22 μs
128 | 46 μs | 687 μs | 40 μs | 32 μs
1024 | 51 μs | 702 μs | 57 μs | 103 μs
10000 | 106 μs | 723 μs | -- | 492 μs
20000 | 134 μs | 380 μs [3] | -- | 866 μs
30000 | 160 μs | 800 μs | -- | 1237 μs 

[1] Only works with input sizes <= 2048
[2] LookupTable with 256 features
[3] thrust::sort is not quite as slow for input sizes close to 20,000 for some reason

I'm happy with putting ModernGPU in cutorch, if that makes more sense.

@wickedfoo, @soumith, thoughts?

dominikgrewe:
I'm concerned about putting a library as big as ModernGPU (~50 files) into cutorch or cunn, just because of its sort function. Can we maybe adapt THCudaTensor_sort to deal with larger tensors? Or could we take ModernGPUs implementation? Not sure what the license implications would be.

soumith:
@dominikgrewe jeff looked at adapting THCudaTensor_sort to deal with larger tensors, and it is a non-trivial amount of work.

If possible (and easy) we can refactor the sort stuff from ModernGPU. The license is fine, they want to just ship with the Copyright message.

If not possible (or easy), we have to think of an alternative.

wickedfoo:
THCudaTensor_sort could be made to work for arbitrarily sized tensors, it
just requires global memory sorting passes for either bitonic or merge sort.

It's not trivial but it's not super hard. Maybe 1-3 days work.

Though ModernGPU would likely be faster, or perhaps we can extract the
relevant algorithms from MGPU and/or Thrust.

THCudaTensor_sort in fact works for the sizes in question here, it just
defers to Thrust.

The problem is that Thrust does not in general work well on any problem
<500k elements or so, it's not designed for small to medium sizes.
THCudaTensor_sort has its own imp for <=2048 and defers to Thrust above
that if the number of slices is small.

colesbury:
Here are some options I've considered:

1) Use thrust::sort. This would slow down backprop considerably for many use cases

2) Use THCudaTensor_sort. This delegates to thrust::sort for input sizes > 2048, so we have some of the same problems.

3) Improve THCudaTensor_sort. I think the best way to do this would actually be to make use of moderngpu's sort routine as it's fast for a wide range of inputs. Currently THCudaTensor_sort doesn't work for a number of inputs.

4) Include moderngpu in cutorch. The sort routine is immediately useful, but I think there is a good chance that we'll make use of other mpgu routines in the future.

5) Extract out the sort routine from moderngpu. This would be a decent amount of work and a large number of functions (a few dozen). I think we'd have many of the same disadvantages of (4), but without the benefit of easy reuse of other mgpu functions.

I think including moderngpu makes the most sense, but if people are opposed to including it, I can use another sort function for now and we can address sorting later. I think even with the slower sort function the CUDA LookupTable be useful to a number of people.

colesbury:
I've removed the moderngpu dependency

soumith:
hurray, can this now get in?

zhangxiangxiao:
This looks awesome!

PR #94, opened 2015-06-08
DataParallelTable cleanup/refactoring.

clementfarabet:
- cleaned up syntax (camelCase)
- got rid of global variables (use th -g to monitor globals);
  there were 2 of them in DataParallelTable
- supporting batches of arbitrary sizes: do not have to be
  multiples of nb of GPUs anymore; added tests for that
  (so 3 GPUs can be used for batches of size 128, and 
   for cases where batch size varies dynamically, everything
   still works)


soumith:
nice! thanks :)

PR #95, opened 2015-06-08
fix MultiMarginCriterion

nicholas-leonard:
fixes a bug introduced [here](https://github.com/torch/cunn/commit/a579aec6eebb71c0bd63c2e7fa30606752913e9e#diff-0bd6b3577d671633a664145fc6172330L169)


soumith:
thanks! sorry about that...

PR #98, opened 2015-06-18
SpatialConvolutionMM accepts padW/padH != 1

fmassa:
Cuda version of https://github.com/torch/nn/pull/293


soumith:
thanks francisco!

PR #99, opened 2015-06-18
reduce SpatialConvolution padW and padH bounds in test

szagoruyko:
after https://github.com/torch/cunn/pull/98 input tensor size could go negative and test would fail

### DIFF
--- a/test.lua
+++ b/test.lua
@@ -1054,8 +1054,8 @@ function cunntest.SpatialConvolutionMM_forward_single()
    local sj = math.random(1,3)
    local outi = math.random(1,64)
    local outj = math.random(1,64)
-   local padW = math.random(0,8)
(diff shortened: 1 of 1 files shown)


fmassa:
Indeed, there is a small probability that some input tensors to have negative values with the default parameters in the tests, and pad=1 is the maximum value one can have to guarantee a positive input dimension (given the current values for the other variables). But is putting the padding to be in the range [0,1] representative enough ? Maybe we should change the other values as well, and increase the pad to, i.e., [0,2] ?

szagoruyko:
it is failing quite often. I'd say [0,1] is enough

fmassa:
@szagoruyko the probability of failure given the current parameters is 5%. I wouldn't say that it happens often...

PR #101, opened 2015-06-20
Add test case for 2D input to LookupTable

colesbury:
There was a bug for handling 2D (batch) input in CUDA LookupTable.
This adds a test for that case.


zhangxiangxiao:
Sorry I am a bit confused. Does this mean pull request #90 broke LookupTable in batch mode?

soumith:
Yep. Briefly.

zhangxiangxiao:
Thanks for the info @soumith !

PR #102, opened 2015-06-21
add #include 'THClApply.cuh' to Tanh.cu, so builds independently of include order

hughperkins:
add #include 'THClApply.cuh' to Tanh.cu, so builds independently of include order, even if other modules are commented out

I know its a pretty minor change, with no benefit to anyone really :-)

(Background, I modified init.cu to comment most stuff out, except pooling, convolution, tanh, and a couple of others, so it builds super-fast :-) )


soumith:
thank you :)

Issue #104, opened 2015-07-03
Non-deterministic LookupTable


tkocisky:
The backward pass of the LookupTable seems to be non-deterministic on GPU, and different to the CPU implementation.  Is this expected?

```lua
require 'torch'                                                                 
require 'cutorch'                                                               
require 'nn'                                                                    
require 'cunn'                                                                  
                                                                                
do                                                                              
  local lt = nn.LookupTable(4096, 256):cuda()                                   
  lt.weight:fill(1)                                                             
  lt.gradWeight:fill(0)                                                         
  local input = torch.CudaTensor(4000):fill(1)                                  
  lt:forward(input)                                                             
  print(lt.output:sum())                                                        
  lt:backward(input, lt.output)                                                 
  print(lt.gradWeight:sum())                                                    
end                                                                             
                                                                                
do                                                                              
  local lt = nn.LookupTable(4096, 256)                                          
  lt.weight:fill(1)                                                             
  lt.gradWeight:fill(0)                                                         
  local input = torch.DoubleTensor(4000):fill(1)                                
  lt:forward(input)                                                             
  print(lt.output:sum())                                                        
  lt:backward(input, lt.output)                                                 
  print(lt.gradWeight:sum())                                                    
end
```

Output:
```
1024000
641312     -- this varies
1024000
1024000
```

nicholas-leonard:
If you use an input of 768 or less it works, otherwise it fails. So the problem seems to be located in this kernel : https://github.com/torch/cunn/blob/master/LookupTable.cu#L79 . @colesbury could you look into this?

colesbury:
Sure, I'll take a look

tkocisky:
Thanks!

PR #105, opened 2015-07-06
Fix incorrect LookupTable backward pass

colesbury:
Fixes issue #104

The gradient computation in LookupTable was incorrect for
tables with more than 128 features when the input was larger
than 768.


colesbury:
(updated cunn test to cover this case)

Issue #53, opened 2014-12-07
SpatialMaxPooling failure case


soumith:
Fails for the case:
```lua
m=nn.SpatialMaxPooling(2,2,2,2):cuda()
input=torch.randn(128,1024,12,12):cuda()
m:forward(input)
```
```
error in SpatialMaxsampling.updateOutput: invalid argument
luajit: ...ch-distro/install/share/lua/5.1/nn/SpatialMaxPooling.lua:18: aborting
```

jonathantompson:
Nice catch.

soumith:
The reason here is that cunn is complied with 2.0 compability:

https://github.com/torch/cunn/blob/master/CMakeLists.txt#L19

If I switch that to 35, no problems

russellfei:
@soumith Same catch here.

```
a = torch.rand(64,1024,12,12):cuda()
b = nn.SpatialMaxPooling(2,2,2,2):cuda():forward(a)
-- or -- 
b = nn.SpatialAdaptiveMaxPooling(8,8):cuda():forward(a)
```
gives the `invalid argument` error.
Maybe I should switch that arch to 35.
Thanks for instructions above.

bamos:
Hi, I have the same issue with `SpatialAveragePooling` and changing the arch to 35 in `CMakeLists` fixes the problem.

PR #106, opened 2015-07-08
Deterministic max-pooling

szagoruyko:
This version is
* Deterministic (according to @mys007)
* Faster (close to cudnn)
* Supports floor and ceil modes
* Supports padW and padH
* Uses 2 times less memory (stores only indexes, not x and y as before) as CPU version

(kernels borrowed from caffe)
CPU version is here https://github.com/torch/nn/pull/309, both PRs are cross-reviewed by me and @fmassa

### DIFF
--- a/SpatialMaxPooling.cu
+++ b/SpatialMaxPooling.cu
@@ -1,173 +1,80 @@
 #include "utils.h"
 
-/*
- * Description:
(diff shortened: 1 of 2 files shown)


fmassa:
BTW, this PR also fixes https://github.com/torch/cunn/issues/53

soumith:
thanks a lot!

PR #107, opened 2015-07-09
Updated required Cuda version in cmake to 6.5

ipanchenko:
Cuda 5.5 is not enough for cunn. I have tested on 7.0 - it works just fine.


dominikgrewe:
This might be a bit harsh on people with a slightly out-of-date CUDA version. I'm not sure we're using any CUDA 7 specific features, so bumping the required CUDA version to 6 or 6.5 might be better.

ipanchenko:
Yes, you right! But I was not sure if someone uses cuda-7.0 features. So I have changed to cuda-6.5. Do you suppose it would be OK?

hughperkins:
As a general rule, you want to put as low a number as possible.  No point in alienating users.

hughperkins:
if youre not sure, and if putting 5.5 is not causing support issues / grief, I'd be tempted just to leave it as is...

ipanchenko:
5.5 is not working. we can probably change to 6.0 and if it will cause an error we can make 1 more patch, is it OK?

hughperkins:
I should prefix with the caveat that I'm neither core nor a CUDA user.  My interest in this is purely that the more users torch has, cuda or not, the better for any projects vaguely related to torch.

With the caveat out of the way, I think it would be interesting to know what functionalities are broken on 5.5.  is it 1 or 2 specific functionalities?  Everything you tried?

ipanchenko:
"require 'cunn' " does not work on cuda 5.5.
Unfortunately, I do not have enough time+expertise to debug it.
Here is the description: https://groups.google.com/forum/#!topic/torch7/WAaDEAB9KV8

hughperkins:
Ah.  My apologies.  I assumed, incorrectly, that you are a developer wanting to add lots of fun 7.0 functionality :-)  Ok I'm glad that you managed to get your system working, and I'm sorry for the pain you had to go through to get it working.

soumith:
probably time!

hughperkins:
Hmmm, yes :-)

Issue #108, opened 2015-07-09
TemporalMaxPooling breaks with >1024 features


jpiersol:
If the output frame size of a cunn TemporalMaxPooling layer is greater than 1024, the output is no longer correct. Compare the CPU and GPU output of the code below:

```lua
require('torch')
require('cutorch')
require('cunn')

-- ConvNet parameters
kW = 20; dW = 1; maxPool = 4;
numFeatures1 = 20


print("Loading data...")
docData = torch.randn(4200, 530) --this data is transposed from the usual data format
nSamples = docData:size()[1]
nExamples = docData:size()[2]
n = nSamples


print("Creating neural network...")

-- ConvNet
n1 = (n-kW)/dW+1
n2 = torch.floor(n1/maxPool)

cpuFeatsNet = nn.Sequential()
tConvLayer = nn.TemporalConvolution(1, numFeatures1, kW)
cpuFeatsNet:add(tConvLayer)
cpuFeatsNet:add(nn.TemporalMaxPooling(maxPool))

gpuFeatsNet = nn.Sequential()
tConvLayerGpu = nn.TemporalConvolution(1, numFeatures1, kW)
gpuFeatsNet:add(tConvLayerGpu)
gpuFeatsNet:add(nn.TemporalMaxPooling(maxPool))
gpuFeatsNet:cuda()

batchSize = 530
numProcSamples = 530

tConvLayer.bias:zero()
tConvLayerGpu.bias:zero()
tConvLayerGpu.weight[{}] = tConvLayer.weight

function saveNewFeatures(krnToSave, procUnit)
  procUnit = procUnit or "cpu"
  
  if procUnit == "cpu" then
    featsNet = cpuFeatsNet
    print("Processing on the CPU...")
  elseif procUnit == "gpu" then
    featsNet = gpuFeatsNet
    cudaInp = torch.CudaTensor(batchSize,nSamples,1)
    print("Processing on the GPU...")
  end
  
  local feats = torch.Tensor(numProcSamples, n2)
  
  for i=1,numProcSamples,batchSize do
    tIdx = i
    inp = docData[{{1,nSamples},{tIdx,tIdx+batchSize-1}}]:t():reshape(batchSize,nSamples,1):double()
    
    if procUnit == "cpu" then
      feats[{{i,i+batchSize-1},{}}] = featsNet:forward(inp)[{{},{},krnToSave}]
    elseif procUnit == "gpu" then
[...]

PR #109, opened 2015-07-10
Fix index error when length is more than 1024 for TemporalMaxPooling

zhangxiangxiao:
This fixes issue #108. The bug was very obvious and simple.

Tested for both forward and backward propagation on inputs of huge lengths. Did not touch test because it may blow up the time, but let me know if you want it there.


soumith:
Thanks mate

PR #110, opened 2015-07-16
parallel build

szagoruyko:


zhangxiangxiao:
I wish there is a 'like' button on github! We had something like this in an internal cunn at NYU a long time ago, but never created a pull request. This makes writing new modules and debugging much much faster!

PReLU.cu
```diff
@@ -2,12 +2,22 @@
...
       i < (n);                                       \
       i += blockDim.x * gridDim.x)
 
+// Use 1024 threads per block, which requires cuda sm_2x or above
+const int CUDA_NUM_THREADS = 1024;
+
+// CUDA: number of blocks for threads.
+inline int GET_BLOCKS(const int N) {
+  return (N + CUDA_NUM_THREADS - 1) / CUDA_NUM_THREADS;
+}
```
dominikgrewe: Instead of duplicating this wherever it's used, it'd be better to put this into a shared header file.

init.cu
```diff
@@ -10,37 +10,6 @@
 
 #include "utils.c"
```
dominikgrewe: While we're at it, "utils.c" should also be compiled separately.

dominikgrewe:
Thanks a lot! This was long overdue :)

Do we really need to make every function global though? Isn't it enough to make the `cunn_*_init` functions global?

szagoruyko:
@zhangxiangxiao agree!
@dominikgrewe you're right, that was an overkill. Will update now.

PReLU.cu
szagoruyko: do you have a name for it? :)

dominikgrewe: common.h ?

szagoruyko: I was thinking about something fancy. Do you think CUDA_NUM_THREADS should be there too?

dominikgrewe: Yes, I think so. In fact, with the current definition of `GET_BLOCKS` it has to be there.

hughperkins:
cool :-)

szagoruyko:
@dominikgrewe updated

dominikgrewe:
Thanks! Is there a reason why you renamed `utils.c` to `utils.cu`? There's just plain C code in there, so calling it `utils.c` makes more sense to me.

szagoruyko:
because ```extern "C"``` has to used to include "lua.h" in .cu files and caused errors in utils.c.
Actually init.cu could be renamed to init.c too, but there are build errors if I do that.

hughperkins:
```
ifdef __cplusplus
    extern"C" {
#endif
```
 ?

szagoruyko:
@hughperkins thanks Hugh, but then header file has c linkage and all the cu files have c++ linkage and we'd have to add extern "C" to all cunn_*_init files.
I think we should leave them in with cu extension.

dominikgrewe:
Sounds good to me. We already have init.cu, so it's at least consistent.

If you squash your commit this is ready for merge.

szagoruyko:
done

dominikgrewe:
Thanks!

PR #112, opened 2015-07-20
Min and Max use THCudaTensor_min and THCudaTensor_max

szagoruyko:
This PR uses ```THCudaTensor_min``` and ```THCudaTensor_max``` allowing nn.Min and nn.Max inputs to be non-contiguous.

Now backward does not allow arbitrary dimension on input while forward does. Any ideas how to make backward with THC?

### DIFF
--- a/Max.cu
+++ b/Max.cu
@@ -1,40 +1,5 @@
 #include "utils.h"
 
-/*
- * Description:
- *    this function finds the max along the innermost dimension
- *    Nd input, (N-1)d output, (N-1)d argmax
- */
-__global__ void max_output(float *input, float *output, float *indices,
-                           long nrows, long ncols)
-{
-  // output offset:
-  long o = threadIdx.x + blockDim.x * blockIdx.x;
-  if (o >= nrows) return;
-
-  // input offset:
-  long i = o * ncols;
-
-  // move pointers
-  input = input + i;
-
-  // compute max:
-  float max = input[0];
-  long argmax = 0;
-  long ii;
-  for (ii=1; ii<ncols; ii++) {
-      float val = input[ii];
-      if (val > max) {
-          max = val;
-          argmax = ii;
-      }
-  }
-
-  // store
-  output[o] = max;
-  indices[o] = argmax+1;
-}
-
 __global__ void max_gradInput(float *input, float *output, float *indices,
                               long nrows, long ncols)
 {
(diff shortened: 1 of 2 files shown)


hughperkins:
Cool :-)  Also good code reuse/factorization.

dominikgrewe:
I think we can do this entirely in Lua once we have a cutorch implementation of [scatter](https://github.com/torch/torch7/blob/master/doc/tensor.md#tensor-scatterdim-index-srcval). I should have a PR for that some time today (thanks to @hughperkins for the cltorch implementation).

szagoruyko:
Nice! what about
```
values,indexes = tensor:max(dim)
```
can ```values,indexes``` be reused?

dominikgrewe:
Yes, e.g. `torch.max(values, indexes, tensor, dim)`.

szagoruyko:
oh yes sure, didn't know it was overloaded too, thanks

szagoruyko:
but CPU versions of min and max do not accept FloatTensor or DoubleTensor for indexes, right? only LongTensor

hughperkins:
I wsa thinking that maybe we should change that: generalize acceptance of DoubleTensor to torch too.

dominikgrewe:
Good point. We could work around that by overloading the `type()` method, i.e. making sure `indexes` stays a LongTensor unless we change the type to `CudaTensor`. It's a bit hacky, but the problem really lies with cutorch not having a CudaLongTensor, rather than torch.min not accepting DoubleTensor etc as indices.

hughperkins:
For example in char-rnn's OneHot.lua, which was created for torch7, the input to OneHot.lua is a ByteTensor, which has to be cast to a LongTensor to feed to `select`: https://github.com/karpathy/char-rnn/blob/master/util/OneHot.lua#L17

```
function OneHot:updateOutput(input)
  local longInput = input:long()
  self.output:copy(self._eye:index(1, longInput))
```

What is the advantage of forcing the user to coerce the index parameter to LongTensor?

soumith:
scatter/gather have landed. maybe I close this PR and we just do it entirely in lua?

szagoruyko:
I don't think we should go the hacky way preserving indices a LongTensor. @hughperkins has a point, we'd rather add other types to min/max or keep C and CUDA code here.

dominikgrewe:
I don't think keeping the indices as LongTensor is hacky, that's why we can overload the `type` function. It does get a bit hacky for CUDA, because there's no CudaLongTensor, but I think that's still better than keeping custom C implementations.

Maybe having other types for min/max would be nice, but LongTensor would still be the natural choice here anyway. A limitation of cutorch shouldn't affect the design of torch/nn if it can be avoided.

hughperkins:
Using LongTensor ensures that all possible values are valid, but in some cases, the user might only want values from 1 to 128, for which a ByteTensor is sufficient.

dominikgrewe:
Sure, I'm not saying there's no use case for other types, but for `nn.Min` and `nn.Max`, LongTensor is the natural choice.

Supporting other types as indices for min/max is therefore orthogonal to this issue.

PR #116, opened 2015-07-23
Remove CUDA-specific nn.Min and nn.Max implementations.

dominikgrewe:


dominikgrewe:
Require torch/nn#326 to be merged.

PR #122, opened 2015-07-29
SpatialDeconvolution

szagoruyko:
This is to track deconvolution implementation (not yet ready to be merged):
- [x] CUDA implemenation
- [x] im2col and col2im moved to separate header file
- [ ] CUDA test against C
- [ ] C nn implementation @fmassa 
- [ ] C tests @fmassa 

Lua side is here https://github.com/szagoruyko/nn/tree/deconv


SpatialDeconvolution.cu
```diff
@@ -0,0 +1,369 @@
...
+  // Inputs
+  THCudaTensor *input = (THCudaTensor *)luaT_checkudata(L, 2, "torch.CudaTensor");
+  THCudaTensor *gradOutput = (THCudaTensor *)luaT_checkudata(L, 3, "torch.CudaTensor");
+
+  // Params
+  int dW = luaT_getfieldcheckint(L, 1, "dW");
+  int dH = luaT_getfieldcheckint(L, 1, "dH");
+  int kW = luaT_getfieldcheckint(L, 1, "kW");
+  int kH = luaT_getfieldcheckint(L, 1, "kH");
+  int nOutputPlane = luaT_getfieldcheckint(L, 1, "nInputPlane");
```
andreaskoepf: What is the idea behind reversing the meaning of the nInputPlane and nOutputPlane? This is not 'compatible' with @fmassa 's tests in https://github.com/fmassa/nn/blob/deconv/test.lua#L1658 (e.g. in 1670, 1671)

hughperkins:
Observation, I know it's not specific to this test in particular, but using 'random' means that sometimes the test might pass, and sometimes fail.  If you add in a `torch.manualSeed(123)`, at the start of the test, then the geometry of the test will be randomized, but repeatable.

SpatialDeconvolution.cu
```diff
@@ -0,0 +1,369 @@
...
+    THCudaBlas_gemm(
+        state,
+        't', 'n',
+        n_, m_, k_,
+        1,
+        THCudaTensor_data(state, ones), k_,
+        THCudaTensor_data(state, bias), k_,
+        1,
+        THCudaTensor_data(state, output_n), n_
+    );
```
andreaskoepf: I have a general question about 'Do Bias after':

Using the following notation (input+output defined as in a classic O = nn.SpatialConvolution():forward(I)):
ic, ih, iw = input channels, height, width
oc, oh, ow = output channels, height, width
kh, kw = filter kernel height, width

I = input of size (ic, ih * iw)
O = output of size (oc, oh * ow)
im2col(I) producing (ic * kh * kw, oh * ow)
W=weights of size (oc, ic * kh * kw)
b= bias of size (oc, 1)

A classic forward convolution might the look like:
O=W * im2col(I) + b * ones(1, oh * ow)

A deconvolution operation could be:
I=col2im(w^T * (O - b * ones(1, oh * ow))

What confuses me is the addition of the bias to the output:
If the bias is generated by the ctor of nn.SpatialConvolution (base class) we have only a bias vector with oc elements (=which corresponds to the number of input(!)-planes for deconvolution). During deconvolution the convolution-output gets the size of the 'original' input (= we produce ic planes).

Due to the swapping of nOutputPlane, nInputPlane in ln. 15 in this code the nOutputPlanes would be ic in my notation and nInputPlanes oc.

The matrix multiplication in ln. 107 reads from the bias k_ x m_ elements (1, ci) .. but if the bias was defined for for a forward convolution this vector should have oc elements - one entry per output channel (feature-map) of the forward conv.

@szagoruyko mentioned that he checked this impl against caffe and it seems they are also adding the bias afterwards: https://github.com/BVLC/caffe/blob/master/src/caffe/layers/deconv_layer.cpp#L36 (@fmassa thnx for the link).

Swapping nInputPlanes and nOutputPlanes + api expecting column-major matrices + deconv becomes quickly more challenging than I had expected.  Probably I just got something wrong. It took me some time to untangle the bits of Deconv. It would be great if somebody could help me a little bit by checking/correcting my 'math'. :-)  thx

PR #129, opened 2015-08-18
Add strides in SpatialAdaptiveMaxPooling to avoid memory copy in forward

fmassa:
Reduces the forward times for non-contiguous Tensors in the non-batched case by avoiding memory copies.


soumith:
nice! thanks.

PR #133, opened 2015-09-01
Optimized Spatial Batch Normalization

egonina:
This patch adds GPU kernels and host functions for optimized spatial batch normalization routines. This gives up to 4x speedup in forward/backward, up to 6x in validation forward and total of up to 2x boost in overall model training performance compared to using the current batch normalization implementation. 

The new optimized path works for specific problem dimensions, for both affine and non-affine updates:
* batchSize <= 32 or power of 2 i.e. 64, 128, 256, 512, 1025
* nFeature <= 1024
* iH/iW powers of 2

It defaults to the original implementation if it doesn't support the problem dimensions. 

This patch is in conjunction to the SpatialBatchNormalization update in torch/nn/SpatialBatchNormalzation.lua.

Adds unit tests for all the new code paths.

cc @clementfarabet

### DIFF
--- a/SpatialBatchNormalization.cu
+++ b/SpatialBatchNormalization.cu
@@ -1,0 +1,919 @@
+#include <stdio.h>
+#include <sys/time.h>
+
+#include "SpatialBatchNormalization.cuh"
+
+/*
+  NOTE: This correctly handles only 3D/4D tensors with:
+  1. nBatch <= 32, and powers of two >=32 (64, 128, 256, 512, 1024)
+  2. nFeature any value <= 1024
+  3. nSpatial (iH x iW) in powers of two
+
+  This check is performed in nn/SpatialBatchNormalization.lua
+*/
+
+/* Helpers */
+inline int getSpatialChunkSize(int nBatch, int nSpatial) {
+   int threadsPerBatchDim = nBatch >= WARP_SIZE ? nBatch : WARP_SIZE;
+
+   return MAX_THREADS/threadsPerBatchDim > nSpatial ? nSpatial : MAX_THREADS/threadsPerBatchDim;
+}
+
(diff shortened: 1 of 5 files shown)


soumith:
We have something superior in all ways that we can port over and make a PR. It is:
- no restrictions on sizes
- only 2 extra buffers
- 10x + faster
- Both SpatialBatchNormalization and BatchNormalization

let me know if there's interest for that, or we rather merge this in.

fmassa:
@soumith is there a way of updating more often `fbcunn` ? Last commit on cuda code dates from 5 months (excluding a bug fix one month ago), and I couldn't find any reference to `BatchNormalization` in the current repo.
Also, putting some pointers on those additions in the main repo could also redirect people to adapt it to main `torch`/`nn`  ? (which could be a smaller effort than recoding everything from scratch)
Maybe this could avoid duplicating efforts on coding the same thing twice ?

egonina:
@soumith great! yes, that would be very useful for us. Also, agreed with @fmassa

szagoruyko:
I will port facebook SpatialBatchNormalization and BatchNormalization to cunn this week

soumith:
@egonina Sorry if my message sounded condescending. Someone on twitter pointed out that the tone of the message was not nice. I was simply writing an informational message, and will be nicer in the future.

@fmassa @egonina Yes, I think we should communicate more. I will make an effort to open an issue as soon as we start implementing a feature so that others dont overlap efforts. It is often frustrating to be in that place.

davek44:
It looks like this effort stalled out at a pull request from @szagoruyko. Is it ready to be included?

egonina:
If the PR from @szagoruyko has been merged and performance looks good we can close this one.

soumith:
Fixed via #185

PR #134, opened 2015-09-06
SpatialAveragePadding padding and ceil kernels

szagoruyko:
cuda part for https://github.com/torch/nn/pull/365

### DIFF
--- a/SpatialAveragePooling.cu
+++ b/SpatialAveragePooling.cu
@@ -1,57 +1,41 @@
 #include "utils.h"
+#include "common.h"
 
-#define CUDA_MAX_THREADS 1024   // this is safe, in reality 256 is our limit
-
-/*
- * Description:
- *    this function avg-pools an input 3D tensor along dimensions 1 and 2
- *    3D input, 3D output
- */
-__global__ void subsample(float *input, float *output,
-                          int input_n, int input_h, int input_w,
-                          int kH, int kW, int dH, int dW)
-{
-  // iterators
-  int xx, yy;
-
-  // output size
-  int output_w = (input_w - kW) / dW + 1;
-  int output_h = (input_h - kH) / dH + 1;
-
-  // compute offsets based on thread/block ID
-  int o = blockIdx.x;
-  int i = o;
-
-  int xx_start = threadIdx.x;
-  int xx_end = output_w;
-  int xx_step = blockDim.x;
-
-  int yy_start = blockDim.y*blockIdx.y + threadIdx.y;
-  int yy_end = output_h;
-  int yy_step = blockDim.y*gridDim.y;
-
-  // select input/output plane
-  output = output + o*output_w*output_h;
-  input = input + i*input_w*input_h;
-
-  // For all output pixels...
-  for(yy = yy_start; yy < yy_end; yy+=yy_step) {
-    for(xx = xx_start; xx < xx_end; xx+=xx_step) {
-      // Compute the mean of the input image...
-      float *ptr_input = input + yy*dH*input_w + xx*dW;
-      float *ptr_output = output + yy*output_w + xx;
-      float sum = 0;
-      int kx, ky;
-      for(ky = 0; ky < kH; ky++) {
-        for(kx = 0; kx < kW; kx++)
-          sum += ptr_input[kx];
-        ptr_input += input_w; // next input line
+template <typename Dtype, bool COUNT_INCLUDE_PAD>
+__global__ void AvePoolForward(const int nthreads,
+    const Dtype* const bottom_data, const int num, const int channels,
+    const int height, const int width, const int pooled_height,
+    const int pooled_width, const int kernel_h, const int kernel_w,
+    const int stride_h, const int stride_w, const int pad_h, const int pad_w,
+    Dtype* const top_data) {
+  CUDA_KERNEL_LOOP(index, nthreads) {
+    const int pw = index % pooled_width;
+    const int ph = (index / pooled_width) % pooled_height;
+    const int c = (index / pooled_width / pooled_height) % channels;
+    const int n = index / pooled_width / pooled_height / channels;
+    int hstart = ph * stride_h - pad_h;
+    int wstart = pw * stride_w - pad_w;
+    int hend = min(hstart + kernel_h, height + pad_h);
+    int wend = min(wstart + kernel_w, width + pad_w);
+    const int pool_size = (hend - hstart) * (wend - wstart);
+    hstart = max(hstart, 0);
+    wstart = max(wstart, 0);
+    hend = min(hend, height);
+    wend = min(wend, width);
+    Dtype aveval = 0;
+    const Dtype* const bottom_slice = bottom_data + (n * channels + c) * height * width;
+    for (int h = hstart; h < hend; ++h) {
+      for (int w = wstart; w < wend; ++w) {
+        aveval += bottom_slice[h * width + w];
       }
-      // Update output
-      *ptr_output = sum/float(kW*kH);
     }
+    if(COUNT_INCLUDE_PAD)
+      top_data[index] = aveval / pool_size;
+    else
+      top_data[index] = aveval / ((hend - hstart) * (wend - wstart));
   }
 }
+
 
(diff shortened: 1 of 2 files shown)


SpatialAveragePooling.cu
```diff
@@ -1,58 +1,42 @@
...
+    Dtype aveval = 0;
+    const Dtype* const bottom_slice = bottom_data + (n * channels + c) * height * width;
+    for (int h = hstart; h < hend; ++h) {
+      for (int w = wstart; w < wend; ++w) {
+        aveval += bottom_slice[h * width + w];
       }
-      // Update output
-      *ptr_output = sum/float(kW*kH);
     }
+    if(COUNT_INCLUDE_PAD)
```
fmassa: I think this condition is inverted. When you include the padding region, you should divide by `pool_size`.

fmassa: Actually, as is both conditions are the same. I think you should replace the `((hend - hstart) * (wend - wstart))` by `kernel_h*kernel_w`.

SpatialAveragePooling.cu
```diff
@@ -133,105 +120,45 @@ static int cunn_SpatialAveragePooling_updateOutput(lua_State *L)
...
+    for (int ph = phstart; ph < phend; ++ph) {
+      for (int pw = pwstart; pw < pwend; ++pw) {
+        // figure out the pooling size
+        int hstart = ph * stride_h - pad_h;
+        int wstart = pw * stride_w - pad_w;
+        int hend = min(hstart + kernel_h, height + pad_h);
+        int wend = min(wstart + kernel_w, width + pad_w);
+        int pool_size = (hend - hstart) * (wend - wstart);
+        if(COUNT_INCLUDE_PAD)
+          gradient += top_diff_slice[ph * pooled_width + pw] / ((hend - hstart) * (wend - wstart));
```
fmassa: Same comment as before, you should replace `((hend - hstart) * (wend - wstart))` by `kernel_h*kernel_w`.

fmassa:
I made some inline coments. Could you add some tests for the new cases ?

hughperkins:
Tests are missing test for `:ceil()` mode?

szagoruyko:
@hughperkins https://github.com/torch/cunn/commit/e0403984d3fb028bdbc96d55351f9ecaa63e5dfe

hughperkins:
Ah :-)  Right :-)

hughperkins:
Hmmm, it looks like this compare with the cpu version?  Does that mean there is a cpu implementation of ceil average pooling to compare with?

fmassa:
@hughperkins yes, there is a CPU implementation, check https://github.com/torch/nn/pull/365
I've written all the tests needed for both PRs to be merged. But there is still a problem with GPU implementation when `count_include_pad` is false, and I couldn't figure out why (in my tests, it seems that `count_include_pad` is always behaving as if it was true, even though we set it to false).

fmassa:
(Note that I'm comparing it against my fork, with the corrections mentioned in this PR https://github.com/torch/cunn/commit/e0403984d3fb028bdbc96d55351f9ecaa63e5dfe )

hughperkins:
(ported to clnn, in branch, https://github.com/hughperkins/clnn/compare/master...avgpool?expand=1 ; pending merge of https://github.com/torch/nn/pull/365 )

fmassa:
I think that after https://github.com/szagoruyko/cunn/pull/2 is merged, this is fine to go.

hughperkins:
Looks like this is ready to be merged right? :-)

fmassa:
@hughperkins @szagoruyko was comparing this implementation against `cudnn` and it seems that the case in which `setCountExcludePad` is true there are some boundary differences. This case is not yet officially present in `cudnn` binding though, and I haven't had a look at it (still using CUDNN R2).

hughperkins:
Can we at least merge in the `nn` version?  Even though this isnt supported in cunn yet, but it is supported in `clnn`, but I cant merge it into `clnn` master until it is present in `nn` master.

fmassa:
The thing is, both `nn` and `cunn` implemetations are equivalent, so if there is a case in which `cunn` does not behave as `cudnn`, then it's also going to be the case for `nn`.
I think what we all want is to have an equivalent implementation everywhere (and as `cudnn` is becoming the standard for CUDA modules, we need to give the same results as them).
I'll try to have a look at `cudnn` R3 version of `setCountExcludePad`, to see what they are actually computing, and I'll come back to you soon (hopefully today).

hughperkins:
Ok... I think for my own usage I'm not using setCountExcludePad.  Perhaps we could split this PR into two: one for ceilmode, and one for setCountExcludePad?  An easy way to do that could be to just add an `assert` that doesnt allow anyone to use setCountExcludePad for now perhaps?  (or to rename it to to `unstable_unsupported_setCountExcludePad` perhaps?)

fmassa:
Hum, yeah, splitting the PR is a possibility... I'll have a look at it today when I have some time.

szagoruyko:
this PR is almost ready to be merged, the only problem is the difference between it and COUNT_EXCLUDE_PADDING mode of cudnn, which we believe to be a bug in cudnn. will update as soon we find out more.

hughperkins:
@szagoruyko Mmmkay.  What if ... can we add something like `assert(COUNT_EXCLUDE_PADDING == 0)` for now?  And then can make a later pull request to enable `COUNT_EXCLUDE_PADDING`?  (I'm happy to submit the appropriate code change to disable COUNT_EXCLUDE_PADDING for now)

SpatialAveragePooling.cu
```diff
@@ -61,67 +45,79 @@ static int cunn_SpatialAveragePooling_updateOutput(lua_State *L)
...
+  else
+  {
+    nInputCols = input->size[3];
+    nInputRows = input->size[2];
+    nInputPlane = input->size[1];
+    batchSize = input->size[0];
+  }
 
-    luaL_argcheck(L, nInputCols >= kW && nInputRows >= kH, 2, "input image smaller than kernel size");
+  luaL_argcheck(L, nInputCols >= kW - padW && nInputRows >= kH - padH, 2, "input image smaller than kernel size");
```
fmassa: I think this should be `nInputCols >= kW - 2 * padW`, and the same thing for `H`.
Currently, this condition doesn't allow the case of `1x1` input to be used with `kW=kH=3` and `padW=padH=1` (which I think should be valid).

SpatialAveragePooling.cu
```diff
@@ -241,78 +181,91 @@ static int cunn_SpatialAveragePooling_updateGradInput(lua_State *L)
...
+  else
+    AvePoolBackward<float, false>
+      <<< GET_BLOCKS(count), CUDA_NUM_THREADS, 0, THCState_getCurrentStream(state) >>> 
+        (count,
+        THCudaTensor_data(state, gradOutput),
+        batchSize, nInputPlane, nInputRows, nInputCols, nOutputRows, nOutputCols,
+        kH, kW, dH, dW, padH, padW,
+        THCudaTensor_data(state, gradInput));
+
+  THCudaTensor_free(state, gradOutput);
```
fmassa: This is being freed twice

SpatialAveragePooling.cu
```diff
@@ -241,78 +181,91 @@ static int cunn_SpatialAveragePooling_updateGradInput(lua_State *L)
...
+  {
+    // ensure that the last pooling starts inside the image
+    // needed to avoid problems in ceil mode
+    if ((nOutputRows - 1)*dH >= nInputRows + padH)
+      --nOutputRows;
+    if ((nOutputCols  - 1)*dW >= nInputCols  + padW)
+      --nOutputCols;
+  }
+
+  gradOutput = THCudaTensor_newContiguous(state, gradOutput);
```
fmassa: Redundant `new_Contiguous` call, has already been called in https://github.com/torch/cunn/pull/134/files#diff-96a3e9b178afa9682543a117c11f623aR194

szagoruyko:
fixed Francisco's comments and squashed

soumith:
Thanks Sergey.

PR #135, opened 2015-09-09
Update SoftMax to work in spatial mode

jhjin:
The module applies softmax in space and works both on a single
vector/matrix (1D/2D) or spatial vector/matrix (3D/4D).


hughperkins:
(merged to clnn https://github.com/hughperkins/clnn/compare/85756...softmax2 )

PR #141, opened 2015-09-20
Added Padding Option to VolumetricConvolution

kmul00:
Padding option wasn't previously applicable to the Volumetric Convolution, though the Cuda kernel has it. This is the Cuda version of it.

Padding can be provided in the same was as for `SpatialConvolution` :

    module = nn.VolumetricConvolution(nInputPlane, nOutputPlane, kT, kW, kH, [dT], [dW], [dH], [padT], [padW], [padH])

Default is `0`

It would be wonderful if someone can do the same for the CPU version `VolumetricConvolution.c` in the `nn` package. This PR depends on [this PR](https://github.com/torch/nn/pull/392) as well for the updated `VolumetrciConvolution.lua` file in the `nn` package.

### DIFF
--- a/VolumetricConvolution.cu
+++ b/VolumetricConvolution.cu
@@ -162,6 +162,9 @@ static int cunn_VolumetricConvolution_updateOutput(lua_State *L) {
   int kH = luaT_getfieldcheckint(L, 1, "kT");
   int nInputPlane = luaT_getfieldcheckint(L, 1, "nInputPlane");
   int nOutputPlane = luaT_getfieldcheckint(L, 1, "nOutputPlane");
+  int padD = luaT_getfieldcheckint(L, 1, "padW");
+  int padW = luaT_getfieldcheckint(L, 1, "padH");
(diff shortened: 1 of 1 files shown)


soumith:
I'll wait on the CPU version to merge this in.

kmul00:
Hi @soumith ,

This PR would create conflict with the recently updated ***VolumetricConvolution*** in the `nn` package because of the following two reasons:

1) `VolumetricConvolution.lua` now **unfolds**  the tensors, as could be seen [here](https://github.com/torch/nn/blob/master/VolumetricConvolution.lua#L83). Thus the `cuda` version also needs to be changed. For example [this line](https://github.com/torch/cunn/blob/master/VolumetricConvolution.cu#L254) will throw an error.

2) `VolumetricConvolution.lua` now calls `VolumetricConvolution_MM` implicitly ([here](https://github.com/torch/cunn/blob/master/VolumetricConvolution.cu#L254)). Thus that also needs to be changed in the `cuda` implementation.

I will send out a PR very soon with the required changes.

PR #143, opened 2015-09-24
Fast SpatialBatchNormalization (Facebook)

szagoruyko:
An amazing work of @nicolasvasilache , fast and no extra memory.
In fact I ported it more than a week ago and couldn't find time to factor out c++11 and sm_35. So to be added:
 - [x] remove c++11
 - [x] put a condition on sm_35 or find a workaround
 - [ ] port test against CPU version
CPU patch and BatchNormalization are coming too.


nicolasvasilache:
Thanks but this is not yet optimized btw ;)
We could get 2-3x better perf if we wanted to spend a bit more time.

soumith:
fixed via #185

Issue #148, opened 2015-09-29
[cmake] compile with the system card's compute capability


soumith:
cmake at compile-time should be able to determine the compute capability of the card. we should use this to set nvcc flags, rather than use sm20 by default.

Will fix https://github.com/torch/cunn/issues/144

albanD:
Note: [this](https://github.com/BVLC/caffe/blob/6eae122a8eb84f8371dde815986cd7524fc4cbaa/cmake/Cuda.cmake#L9-L132) is the code in Caffe that is doing exactly what you're looking for I think.

hughperkins:
Ah, nice find :-)  I tried dabbling with `nvidia-smi` options, and `nvcc` options, but nothing obvious popped out.

albanD:
@soumith  I think this ca

PR #151, opened 2015-10-07
Add automatic CUDA architecture detection

albanD:
Following the discussion in #148 , here is an automatic detection scheme for the CUDA architecture inspired by the one in Caffe.

Supports:
* Multi-gpu with different architectures
* If it fails, will compile for all known architectures (the list is [here](https://github.com/Moodstocks/cunn/blob/f3facac05ca9b51fa8d8ed9e25979fe107a4da17/cmake/FindCudaArch.cmake#L6) and should be updated as new architectures goes out)
* You can specify with the cmake flag `CUDA_ARCH_NAME` to compile
    * only for the detected architecture (`Auto`, this is the default)
    * all architectures (`All`, this is used if there is an error in `Auto` mode)
    * a specific GPU type (`Fermi`, `Kepler` or `Maxwell`, see [here](https://github.com/Moodstocks/cunn/blob/f3facac05ca9b51fa8d8ed9e25979fe107a4da17/cmake/FindCudaArch.cmake#L97-L102) for what is used for each of these types)
    * a specific architecture (`Manual`) and specify the architecture with `CUDA_ARCH_BIN` (for example to compile only for compute capability 5.2: `-DCUDA_ARCH_NAME=Manual -DCUDA_ARCH_BIN=5.2`)

I have the same thing for cutorch prepared. I will wait for the feedback on this one before opening a PR in the cutorch repo.


soumith:
this is awesome! will review later today.

jonathantompson:
Do you get better performance if you target a newer architecture?

soumith:
Possibly (?), because instructions like __shfl or __ldg are not compiled down to their backward-compatible slower equivalents in bytecode, but the instruction is retained as is.

albanD:
A quick test using [this](https://github.com/Moodstocks/gtsrb.torch) repo (`th main.lua -n 5000`), on a GTX 970 comparing sm_20 and sm_52 (CUDA 7.5 and cudnn R3) leads to no clear improvement.
If you look very closely you go from 1113MB to 1111MB memory usage and the time difference is smaller than the 0.03 ms per sample which is the difference I have on different runs with the same architecture (takes ~5.21 ms to process a sample).

fmassa:
@albanD I think that using `cudnn` masks the possible benefits, because `cudnn` is not affected by this CMake change.
But maybe native `cutorch` functions (as well as `cunn`s) could get some improvement ?

soumith:
A good perf check would be on a function that uses __shfl or __ldg. you can search in cutorch for those.

albanD:
@fmassa the repo with the command I putted does not use cudnn.
If you want to use cudnn you can do `th main.lua -n 5000 --net use_cudnn.lua`. FYI this leads to 1121MB memory use and ~4.38 ms to process a sample

fmassa:
@albanD apologies, your previous post mentioned `cudnn R3` so I was mislead by that.

jonathantompson:
Awesome thanks for the update.

I have noticed that extremely large matrix operations are not supported on
sm_20 (because of max workgroup size limitations), so I've already switched
all my CMakeLists.txt to the new sm...  So this change is extremely
welcome!  Thanks @albanD!

soumith:
let's see what happens :)

szagoruyko:
good job thanks @albanD and moodstocks!

PR #156, opened 2015-10-25
Added LeakyReLU to cunn

rotmanmi:


soumith:
merged via https://github.com/torch/cunn/pull/172

thanks!

PR #158, opened 2015-10-27
removing syncs from DPT, adding flattenParams

soumith:
A patch from Natalia Gimelshein of NVIDIA, which makes DPT way more efficient.

### DIFF
--- a/DataParallelTable.lua
+++ b/DataParallelTable.lua
@@ -185,7 +185,7 @@ end
 local DataParallelTable, parent = torch.class('nn.DataParallelTable',
 'nn.Container')
 
-function DataParallelTable:__init(dimension)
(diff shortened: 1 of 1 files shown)


DataParallelTable.lua
```diff
@@ -198,8 +198,12 @@ function DataParallelTable:__init(dimension)
    self.gradOutputGpu = {} -- gradOutputs for each gpu
    self.outputGpu = {} -- outputs for each gpu
    self.gradInputGpu = {} -- gradInput for each gpu
+   self.flattenParams = true
```
fmassa: Is this supposed to be always true ? If yes  we can take it off from the constructor, right ?

soumith: will-fix

PR #159, opened 2015-10-28
Added MarginCriterion with unit test

htwaijry:
Hi,

Adding MarginCriterion cuda code with unit test.
I'm not very familiar with cuda, but the code was based off "MultiMarginCriterion.cu". It passes the 
unit test comparing against the non-cuda version.

-Hani


soumith:
Thanks a lot Hani!

PR #166, opened 2015-11-21
Allow ceil_mode in SpatialAveragePooling

jhjin:


szagoruyko:
there is https://github.com/torch/cunn/pull/134 already

jhjin:
Oh I missed those PRs.
https://github.com/torch/nn/pull/365
https://github.com/torch/cunn/pull/134

Then I will wait until they are merged. Thanks for pointing it out.

PR #168, opened 2015-11-26
VolumetricConvolution now supports padding

kmul00:
***VolumetricConvolution*** now supports *padding*.

`cuda` version of [this](https://github.com/torch/nn/pull/481) change in `nn`.

### DIFF
--- a/VolumetricConvolutionMM.cu
+++ b/VolumetricConvolutionMM.cu
@@ -1,0 +1,527 @@
+#include "utils.h"
+#include "common.h"
+
+// Kernel for fast unfold+copy
+// Borrowed from Theano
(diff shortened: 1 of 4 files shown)


fmassa:
I don't think that copying the old `VolumetricConvolution` with a different name and minor changes is a good strategy.
Couldn't you just change the lua file to handle differently Cuda and C cases ? Something like
```lua
if input:type() == 'torch.CudaTensor' then
-- just call the previous function
else
-- do the reshaping etc and call the MM function
end
```

kmul00:
Yeah, that would be better. I will do it then. Then the already `cuda` version is also good to go. I can close this PR then.

PR #173, opened 2015-12-03
add nccl support to DataParallelTable

ngimel:


soumith:
this is super cool. Thanks Natalia.

PR #175, opened 2015-12-04
Decrease required precision for SmoothL1, mse and Tanh tests.

dominikgrewe:
These were failing regularly when running the tests repeatedly on my machine.


soumith:
Thanks!

PR #176, opened 2015-12-06
fixing FullConvolution tests

### LINKED ISSUE #174: SpatialFullConvolution test flaky
The SpatialFullConvolution tests sometimes fail due to invalid configurations:

```
SpatialFullConvolution_forward_batch
 Function call failed 
.../cunn/test.lua:785: bad argument #1 to 'forward' (3D or 4D (batch mode) tensor is expected)
```

It can be reproduced when running
```sh
torch -lcunn -e "nn.testcuda({'SpatialFullConvolution_forward_batch','SpatialFullConvolution_backward_batch'}, false, 1, 1449228794)"
```

The input sizes don't make sense: `16,14, 0.33333333333333, 2.3333333333333`

I've only observed this in cunn, but I suspect the same could happen in nn.

soumith:
fixes #174

### DIFF
--- a/test.lua
+++ b/test.lua
@@ -723,8 +723,8 @@ function cunntest.SpatialFullConvolution_forward_single()
    local padH = math.random(0,1)
    local outi = math.random((ki - si - 2 * padW) + 1, 64)
    local outj = math.random((kj - sj - 2 * padH) + 1, 64)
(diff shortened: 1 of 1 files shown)


dominikgrewe:
Thanks Soumith!

PR #177, opened 2015-12-16
THCUNN: Functional C backend conversion using ffi

andreaskoepf:
This PR is the cunn counterpart of https://github.com/torch/nn/pull/524.

The modules Abs and AbsCriterion have been moved into a separate libTHCUNN. The module-functions no longer have a single lua_State* parameter but a full signature, e.g. void THNN_(Abs_updateOutput)(THNNState *state, THTensor *input, THTensor *output);.

Please see the [nn PR](https://github.com/torch/nn/pull/524) for more details.


soumith:
rebased and merged

PR #178, opened 2015-12-17
Fix SpatialMaxPooling (check if last pool block starts in image)

clement-masson:
Add a check (that is present in the C version of SpatialMaxPooling but was not in CUDA version) so that the last pooling blocks do not start out of the image.

It was the cause of some errors in nn.testcuda (defined in cunn/test.lua)


fmassa:
Thanks @clement-masson , that was indeed missing in the CUDA version.

soumith:
Thanks clement.

PR #182, opened 2015-12-30
THCUNN functional module conversions batch 1

andreaskoepf:
- [x] ClassNLLCriterion
- [x] DistKLDivCriterion
- [x] HardTanh
- [x] L1Cost
- [x] final cross-check test with old implementation


andreaskoepf:
Cross-checked with old impl via [xtest.lua](https://github.com/andreaskoepf/cunn/blob/cunn2/xtest.lua).

PR #183, opened 2015-12-31
GPU implementation for locally connected layers

yangky11:
This PR is the corresponding GPU implementation for https://github.com/torch/nn/pull/546.
I'm still working on tests.


yangky11:
The new layer passed these tests (on Tesla K40 and Geforce GTX Titan X).
```
th -lcunn -e "nn.testcuda{'SpatialConvolutionLocal_forward_single',   'SpatialConvolutionLocal_forward_batch', 'SpatialConvolutionLocal_backward_single',   'SpatialConvolutionLocal_backward_batch'}"  
```
Ready to be merged.

soumith:
One last but important comment / change needed. Can you revert the badd*mm calls back to addmm, baddmm doesn't work well with multi-GPU yet because it uses scratch space that's alloc-ed on every call, and so there's a force synchronize that happens for every call to baddmm calls, and multi-GPU asynchronous dispatch doesn't work.

soumith:
@yangky thanks a lot for this contribution!

yangky11:
@soumith  This is not ready for merge yet. My GPU server has some issues at this moment and I will complete this PR when it's working again.

PR #185, opened 2016-01-05
Add SpatialBatchNormalization CUDA implementation

colesbury:
Adapted from @nicolasvasilache's implementation and @szagoruyko's port.

This uses less memory than the existing fbcunn implementation and on my ResNet benchmarks is slightly faster than both the current fbcunn and the cudnn implementation.

Here's a little benchmark:
https://gist.github.com/colesbury/51991501334c9e4b1728


soumith:
Thanks Sam.

nicolasvasilache:
excellent!

PR #186, opened 2016-01-06
Add SpatialMaxUnpooling

clement-masson:
Hi all.
This is the cuda version of SpatialMaxUnpooling previously proposed in torch/nn.
(https://github.com/torch/nn/pull/537)

Best,
Clément

### DIFF
--- a/SpatialMaxUnpooling.cu
+++ b/SpatialMaxUnpooling.cu
@@ -1,0 +1,146 @@
+#include "utils.h"
+#include "common.h"
+
+template <typename Dtype>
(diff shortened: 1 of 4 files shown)


soumith:
Thanks Clement!

szagoruyko:
the dedicated implementation for unpooling will be removed in THNN as `updateOutput`/`updateGradInput` are essentially `updateGradInput`/`updateOutput` from `SpatialMaxPooling`, @fmassa will do it soon
Same is probably possible with `SpatialFullConvolution`

PR #188, opened 2016-01-08
Fix SpatialBatchNormalization:accGradParameters()

colesbury:
When the CUDA backwards implementation is called by accGradParameters
gradInput is nil/empty. Use the size from gradOutput instead.

I also added tests for both backward() and updateGradInput() followed by
accGradParameters() to test.lua.


soumith:
Thanks Sam

PR #189, opened 2016-01-11
Add multi-threading support in DataParallelTable

colesbury:
This lets us use the fastest CUDNN R4 kernels with DPT, which have much
higher launch overhead. With single-threaded dispatch, the launch
overhead from these kernels delays the second, third, and fourth GPUs so
much that it actually increases the total time.

I've also included some other optimizations:
- `gradInput` can be set to `nil` when the DataParallelTable is the
  top-level module; this saves time and memory
- input can optionally be on the CPU; this saves memory on the first
  GPU, and is faster when the input is in pinned memory since the copies
  to each GPU happen concurrently.

This change is mostly backwards compatible, with the below caveats:
- The preferred way to add modules is by a single call to add with the
  list of GPUs i.e. `:add(module, {1, 2, 3, 4})`
- `DataParallelTable:get()` only works with the first module

With these changes I'm able to realize a 20-25% speedup with MSRA's ResNet model on 4 GPUs, mostly because it allows the use of the fastest CUDNN R4 kernels.

@dominikgrewe, if you have time, your review would be appreciated

### DIFF
--- a/DataParallelTable.lua
+++ b/DataParallelTable.lua
@@ -1,189 +1,45 @@
-local gpuLocalCopyBuffers = {}
-local baseGpuIndex = 1  -- A constant
+--[[
+   This file implements data parallelism for Torch modules.
 
--- *****************************************************************************
--- Helper Functions
--- *****************************************************************************
--- queryGPUDeviceId - Function to query a tensor or table for the
--- GPUID.  For tables we will search the table for CudaTensors, query their
--- device and make sure the deviceIds of ALL CudaTensors are on the same GPU.
-local function queryGPUDeviceId(object)
-   if torch.type(object) == 'torch.CudaTensor' then
-      return object:getDevice()
+   The same model is replicated on multiple GPUs. The input is split, typically
+   into smaller mini-batches. Each replicated model handles only its portion of the input.
+   The weight updates for each replica are summed together on the first replica
+   in accGradParameters.
+
+   By default, this module uses only one thread and relies on asynchronous kernel launches.
+   To use multiple threads, call DataParallelTable:threads(initFunc).
+
+   For best performance, install NCCL:
+    https://github.com/NVIDIA/nccl
+    https://github.com/ngimel/nccl.torch
+]]--
+local DataParallelTable, parent = torch.class('nn.DataParallelTable', 'nn.Container')
+
+local Impls = {}
+local BasicImpl = torch.class('nn.DataParallelTable.Basic', Impls)
+local ThreadsImpl = torch.class('nn.DataParallelTable.Threads', Impls)
+
+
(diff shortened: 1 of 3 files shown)


DataParallelTable.lua
```diff
@@ -494,106 +371,318 @@ function DataParallelTable:name()
...
+      else
+         t[k] = v
+      end
+   end
+   file:writeObject(t)
+   -- Force synchronization, this keeps you honest
+   self:syncParameters()
+end
+
+function DataParallelTable:sanitize()
```
fmassa: This function should be called [clearStates](https://github.com/torch/nn/pull/526)
Also, I think it should return `self`

colesbury: Ooops! I meant to remove this

colesbury:
This depends on https://github.com/torch/nn/pull/563

DataParallelTable.lua
```diff
@@ -494,106 +371,309 @@ function DataParallelTable:name()
...
+   dst = torch.type(dst) == 'torch.CudaTensor' and dst or torch.CudaTensor()
 
-      asyncGPUCopy(dst, slice)  -- dst, src
+   local index, size = sliceRange(src:size(self.dimension), idx, n)
+   if size == 0 then
+      dst:resize(0)
    else
-      error('input must be a nested table of tensors!')
+      local slice = src:narrow(self.dimension, index, size)
+      dst:resize(slice:size()):copyAsync(slice)
```
ngimel: If copies are made from the GPU, and the stream is non-default, additional sync is required. Torch runs D2D copies on the source device, and destination devices will start calculating before copies are finished. You can profile attached file to observe this behaviour. 
[copytest.txt](https://github.com/torch/cunn/files/96563/copytest.txt)
copytest

colesbury: Ah, good catch! This is true of all the :copy calls (not just copyAsync), right?

I can add the appropriate cutorch.streamWaitForMultiDevice calls

ngimel: Yes, that's true for all D2D non-default-stream copy calls, so deepTensorsAdd and deepTensorsCopy are affected too.

________________________________

ngimel: Yes, that's true for all D2D non-default-stream copy calls, so deepTensorsAdd and deepTensorsCopy are affected too. (Sorry for my botched direct e-mail response).

colesbury: Running DataParallelTable in a non-default stream is fixed now. I added a test for it in test_DataParallelTable.lua

soumith:
@colesbury is this ready for merge?

colesbury:
This is ready for merge

soumith:
Thanks!

PR #190, opened 2016-01-12
Add adjustment term to SpatialFullConvolution to control output size

fbesse:
That mirrors the changes in https://github.com/torch/nn/pull/552


soumith:
Thanks!

Issue #191, opened 2016-01-13
Implementing SoftShrink in GPU


edouardoyallon:
Hi,

Do you think it would be possible to implement SoftShrink for GPU? I'll try on my side, yet I was stated it is simple but I don't think I know enough of Torch and CUDA to do it correctly. 

Thanks a lot.

fmassa:
I'm preparing a patch for it. Will send a PR in a few minutes

soumith:
Fixed in latest trunk!

PR #192, opened 2016-01-13
Add GPU version of SoftShrink

fmassa:
Add missing GPU implementation of SoftShrink, as pointed out in https://github.com/torch/cunn/issues/191


soumith:
Wow so quick! Thanks

PR #193, opened 2016-01-14
Update LogSoftMax to work in spatial domain

fmassa:
Cuda version of https://github.com/torch/nn/pull/560

**This is a hacky PR**.
As I don't have experience with CUDA, I preferred not to touch in the current implementation of `LogSoftMax` (to avoid breaking something). I simply copied the [old kernels of `LogSoftMax`](https://github.com/torch/cunn/blob/6992815706dc39ba3dfbb7e91cc2b76906458981/LogSoftMax.cu), which were simpler to understand and added the strides to handle spatial data, similarly to https://github.com/torch/cunn/pull/135.

Should address https://github.com/torch/nn/issues/548

### DIFF
--- a/lib/THCUNN/LogSoftMax.cu
+++ b/lib/THCUNN/LogSoftMax.cu
@@ -1,5 +1,104 @@
 #include "THCUNN.h"
 
+#define LOGSOFTMAX_THREADS 128
+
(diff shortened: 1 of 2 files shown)


soumith:
See https://github.com/torch/cunn/pull/338

PR #205, opened 2016-02-08
Fix thread configuration in inference mode

ivankreso:
```
      dim3 blocks(input.getSize(1), input.getSize(0));
      dim3 threads(input.getSize(3),
        min(input.getSize(2), maxThreadsPerBlock / input.getSize(3)));

      SpatialBatchNormalizationUpdateOutputInference_kernel
        <<<blocks, threads, 0, s>>>
        (input, output, runningMean, runningVar, weight, bias, eps);
```
Current way of determining thread configuration in inference mode crashes CUDA if input.getSIze(3) i.e. the width of the image is > 1024.
maxThreadsPerBlock is 1024 for modern GPUs and in that case we get:
```
min(input.getSize(2), maxThreadsPerBlock / input.getSize(3)) =  0
```
I modifed the implementation and now it supports all input dimensions in a same way
train version does.

### DIFF
--- a/SpatialBatchNormalization.cu
+++ b/SpatialBatchNormalization.cu
@@ -121,6 +121,7 @@ static THCDeviceTensor<float, Dim> checktensor(lua_State* L, int index) {
   return toDeviceTensor<float, Dim>(getCutorchState(L), t);
 }
 
+template<int NumThreads>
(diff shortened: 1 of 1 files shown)


colesbury:
LGTM, thanks for the fix

PR #211, opened 2016-02-15
Instantiate impl field in older models

apaszke:
Currently, when an old model checkpoint is loaded, it's `impl` field isn't initialized, so it crashes as soon as you try to do anything with the network.


soumith:
Thanks!

PR #212, opened 2016-02-17
Zero gradients on secondary GPUs after accGradParameters

### LINKED ISSUE #209: Reduce gradients problem of DataParallelTable:accGradParameters()
At [DataParallelTable:__backward()#L186](https://github.com/torch/cunn/blob/97f66e19a3b9ce30e4717164b8eaea257b569bf8/DataParallelTable.lua#L186)
```DataParallelTable:accGradParameters()``` calls ```DataParallelTable:__backward()``` and it sums gradients to the base GPU more than once if a user does not run ```nn.Module:zeroGradParameters()``` at every iteration.
In my opinion, gradients should be reduced only once after a user called ```DataParallelTable:backward()``` or ```DataParallelTable:accGradParameters()``` several times.

Any ideas are welcome.

colesbury:
Before this change, calling forward & backward multiple times without calling
zeroGradParameters would compute an incorrect gradient.

Fixes #209

### DIFF
--- a/DataParallelTable.lua
+++ b/DataParallelTable.lua
@@ -216,13 +216,21 @@ function DataParallelTable:__backward(method, input, gradOutput, scale)
    end
 
    if method == 'backward' or method == 'accGradParameters' then
+      local params = self:moduleParameters()
       -- Accumulate the gradients onto the base GPU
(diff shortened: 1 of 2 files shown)


soumith:
Thanks!

Issue #213, opened 2016-02-18
Use DataParallelTable on a deeper network produces different values


Jerrynet:
Use DataParallelTable on a deeper(or complex) network produces different values (compared to single GPU version), which may slow down the converge rate?

I did an experiment on [facebook residual net](https://github.com/facebook/fb.resnet.torch): 
**nn.DataParallelTable() vs. nn.Sequential():add(nn.DataParallelTable())**
Train resnet-50 with following command:
(these are correct outputs, without any modifications of code.)
```
jerry$ torch main.lua -depth 50 -batchSize 32 -nGPU 2 -nThreads 8 -data /data/ILSVRC
=> Creating model from file: models/resnet.lua
 | ResNet-50 ImageNet
=> Training epoch # 1
 | Epoch: [1][1/40037]    Time 24.161  Data 12.260  Err 7.1492  top1 100.000  top5 100.000
 | Epoch: [1][2/40037]    Time 0.236  Data 0.002  Err 30.0546  top1 100.000  top5 100.000
 | Epoch: [1][3/40037]    Time 0.203  Data 0.001  Err 19.9015  top1 100.000  top5 100.000
 | Epoch: [1][4/40037]    Time 0.354  Data 0.001  Err 15.3969  top1 100.000  top5 100.000
 | Epoch: [1][5/40037]    Time 0.219  Data 0.001  Err 15.3319  top1 100.000  top5 100.000
 | Epoch: [1][6/40037]    Time 0.333  Data 0.001  Err 16.4915  top1 100.000  top5 100.000
 | Epoch: [1][7/40037]    Time 0.292  Data 0.001  Err 12.7519  top1 100.000  top5 100.000
 | Epoch: [1][8/40037]    Time 0.548  Data 0.001  Err 9.5735  top1 100.000  top5 100.000
 | Epoch: [1][9/40037]    Time 0.192  Data 0.001  Err 10.5805  top1 100.000  top5 100.000
 | Epoch: [1][10/40037]    Time 0.326  Data 0.001  Err 9.3723  top1 100.000  top5 100.000
 | Epoch: [1][11/40037]    Time 0.495  Data 0.001  Err 7.8360  top1 100.000  top5  96.875
 | Epoch: [1][12/40037]    Time 0.433  Data 0.017  Err 8.4132  top1 100.000  top5 100.000
 | Epoch: [1][13/40037]    Time 0.530  Data 0.080  Err 7.6863  top1 100.000  top5  96.875
[...]

PR #214, opened 2016-02-19
Do flattenParameters() only once

Jerrynet:
Prevent parameters to be flattened more than once, as described in #213


soumith:
cc: @colesbury for review.

colesbury:
I think this is good. There are some cases such as with sharing where it may not work perfectly, but that's going to be hard to avoid. (For example, getParameters doesn't guarantee that p[1]:storageOffset is at the beginning or that p[#p] is at the end).

Jerrynet:
Thanks @colesbury and @soumith.

PR #215, opened 2016-02-19
Fix sizeAverage bug in MultiMarginCriterion (divide by nbatch, not ndim)

soumith:
Original author @adamlerer


soumith:
cuda fix of https://github.com/torch/nn/pull/646

PR #222, opened 2016-02-22
Fixed VolumetricFullConvolution.

fbesse:
The module didn't pass gradients check and had limitations (batch mode only, constraints on kernel size and padding). This new implementation removes these limitations, and is now implemented like SpatialFullConvolution.


fbesse:
NN PR: https://github.com/torch/nn/pull/659

PR #223, opened 2016-02-23
Adding SparseLinear with CUDA

ebetica:
Adding SparseLinear with CUDA. Most of the functions are directly converted from SparseLinear.c. Depending on how well THCudaBlas operations are pipelined, it may be more efficient to write custom kernels for most of the operations. UpdateOutput uses cusparse.


lib/THCUNN/SparseLinear.cu
```diff
@@ -0,0 +1,243 @@
...
+  init_cusparse();
+  float one = 1;
+  cusparseSgemvi_bufferSize(cusparse_handle,
+          CUSPARSE_OPERATION_TRANSPOSE,
+          inDim,
+          outDim,
+          nnz,
+          &pBufferSize
+  );
+  cudaMalloc((void**)&pBuffer, pBufferSize);
```
soumith: could you avoid cudaMalloc and free, and instead preallocate a buffer that is passed-in.
cudaFree causes a device synchronization, which stops us from doing multi-GPU

ebetica: The buffer size is not known ahead of time... I'm not sure how I would preallocate it. Would using a THCudaStorage work?

soumith: I think since this is part of an nn layer, you can initialize it in nn, keep it around and pass it in. You can call THCudaTensor_resize() which will just resize if it needs a bigger buffer.
A similar line of code would be:
https://github.com/torch/nn/blob/master/SpatialConvolution.lua#L51
https://github.com/torch/nn/blob/master/SpatialConvolution.lua#L109
https://github.com/torch/nn/blob/master/SpatialConvolution.lua#L180

and follow the "columns" variable in here:
https://github.com/torch/cunn/blob/master/lib/THCUNN/SpatialConvolutionMM.cu

ebetica:
Should be fixed, you merged the addition in nn half an hour ago Soumith, thanks.

ebetica:
This is now updated with the batch version of sparse linear given in this [commit ](https://github.com/torch/nn/commit/bcdda102686713037538d770ee5c033879450ee3)

lib/THCUNN/SparseLinear.cu
```diff
@@ -0,0 +1,284 @@
...
+
+  colinds = THCudaTensor_newSelect(state, input, 2, 0);
+  colinds_int = THCudaIntTensor_newWithSize1d(state, nnz*batchnum);
+  THCudaIntTensor_copyCudaFloat(state, colinds_int, colinds);
+
+  // See http://docs.nvidia.com/cuda/cusparse/index.html#compressed-sparse-row-format-csr
+  csr_int = THCudaIntTensor_newWithSize1d(state, batchnum+1);
+  init_cusparse();
+  for (h = 0; h < batchnum+1; h++) {
+    THCudaIntTensor_set1d(state, csr_int, h, 1 + nnz * h);
```
soumith: make this for loop a simple-stupid CUDA kernel.

lib/THCUNN/SparseLinear.cu
```diff
@@ -0,0 +1,284 @@
...
+      "gradWeight must be contiguous to use cuda");
+
+  if (!THCudaTensor_isContiguous(state, gradWeight))
+    gradWeight = THCudaTensor_newContiguous(state, gradWeight);
+  THCudaTensor_zero(state, gradBias);
+
+  long batchnum = THCudaTensor_size(state, lastInput, 0);
+  long nnz = THCudaTensor_size(state, lastInput, 1);
+
+  for (h = 0; h < batchnum; h++) {
```
soumith: A cuda module really shouldn't have a double for loop like this. move this whole operation into a CUDA kernel as well. Ask @wickedfoo if any of his generic apply kernels can be leveraged so that you dont need to write this from scratch.

lib/THCUNN/SparseLinear.cu
```diff
@@ -0,0 +1,284 @@
...
+  THArgCheck(THCudaTensor_isContiguous(state, gradOutput), 3,
+             "gradOutput must be contiguous");
+
+  long batchnum = THCudaTensor_size(state, input, 0);
+  long nnz = THCudaTensor_size(state, input, 1);
+  THCudaTensor_resize2d(state, gradOutput, batchnum, outDim);
+
+  for (h = 0; h < batchnum; h++) {
+    for (i = 0; i < nnz; i++) {
+      float val = scale*THCudaTensor_get3d(state, input, h, i, 1);
```
soumith: this is a blocking operation as well. 
let's discuss this PR offline. a bit more work has to be done to remove bottlenecks.

ebetica:
This has been updated to work with the PR @ https://github.com/torch/nn/pull/698

lib/THCUNN/SparseLinear.cu
```diff
@@ -0,0 +1,258 @@
...
+  return t->nDimension == 1 && t->size[0] == size0;
+}
+
+void printCuda(THCState *state, THCudaTensor *tensor, char* str) {
+  thrust::device_ptr<float> ptr(THCudaTensor_data(state, tensor));
+  printf("%s\n", str);
+  thrust::copy(ptr, ptr+THCudaTensor_nElement(state, tensor), std::ostream_iterator<float>(std::cout, "\t"));
+  printf("\n");
+}
+void printCuda(THCState *state, THCudaIntTensor *tensor, char* str) {
```
soumith: This function seems to be double-declared here

lib/THCUNN/SparseLinear.cu
```diff
@@ -0,0 +1,258 @@
...
+  thrust::device_ptr<int> ptr(THCudaIntTensor_data(state, tensor));
+  printf("%s\n", str);
+  thrust::copy(ptr, ptr+THCudaIntTensor_nElement(state, tensor), std::ostream_iterator<float>(std::cout, "\t"));
+  printf("\n");
+}
+
+static THCudaTensor *values = 0;
```
soumith: this seems like a terrible idea (and it's also not thread-safe btw), let's dicuss offline why you're trying to keep state in this file, i think you can remove it.

lib/THCUNN/SparseLinear.cu
```diff
@@ -0,0 +1,258 @@
...
+      THCudaIntTensor_data(state, csr_int),
+      THCudaIntTensor_data(state, colinds_int),
+      THCudaTensor_data(state, nv),
+      THCudaIntTensor_data(state, rowInd),
+      THCudaIntTensor_data(state, colptr),
+      CUSPARSE_ACTION_NUMERIC,
+      CUSPARSE_INDEX_BASE_ONE);
+
+  float one = 1;
+  cusparseMatDescr_t descr=0;
```
soumith: nit: spacing

ebetica:
Fixed nits

soumith:
Thanks Zeming!

PR #234, opened 2016-03-08
Reuse headers from THCUNN.h in THCUNN.lua

fmassa:
Cuda counterpart of https://github.com/torch/nn/pull/674


soumith:
Thanks Francisco!

PR #239, opened 2016-03-22
make margin parameterizable

caldweln:
make margin parameterizable in cunn.MultiMarginCriterion

prevents incorrect arg count error when using MultiMarginCriterion on CUDA device


soumith:
Thanks @caldweln

PR #240, opened 2016-03-28
[LookupTable] Add Max-norm constraints to LookupTable

xianjiec:
The cuda version for the pull request in nn:
 https://github.com/torch/nn/pull/739

### DIFF
--- a/lib/THCUNN/LookupTable.cu
+++ b/lib/THCUNN/LookupTable.cu
@@ -4,9 +4,11 @@
 #include <thrust/device_ptr.h>
 #include <thrust/execution_policy.h>
 #include <thrust/iterator/constant_iterator.h>
+#include <thrust/transform_reduce.h>
 #if CUDA_VERSION >= 7000
 #include <thrust/system/cuda/execution_policy.h>
 #endif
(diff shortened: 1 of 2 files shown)


colesbury:
This LGTM. @xianjiec how does performance with maxNorm on compare to maxNorm off?

xianjiec:
Hey @colesbury, the performance appears to depend on the number of unique indices in the "input". When most of the rows of the "module.weight" need to be re-normed, the running time would increase significantly.

apaszke:
I have only a small nit: how about renaming `value` to something more meaningful like `normType` or simply `p`?

xianjiec:
Hey @apaszke, the variable name is borrowed from torch.renorm: https://github.com/torch/torch7/blob/ad65230b55b832d8e9cf49abf512b9a03f52c0ca/lib/TH/generic/THTensorMath.c#L2299

Should we use a different name here?

apaszke:
I see. Well, I think that it would be better to break out from this convention. It's a very uninformative name.

xianjiec:
The `value` has been changed to `normType` in this pull request and also torch/nn#739.

PR #244, opened 2016-04-08
Add alias cunn.test(), following eg format of cutorch.test()

hughperkins:
Add alias cunn.test(), following eg format of cutorch.test()

(Cos, so many times I type like:

```
luajit -l cunn -e 'cunn.test()'
-- hmmm, nope...
luajit -l cunn -e 'nn.cudatest()'
-- hmmm....
vim ~/torch/install/share/lua/5.1/cunn/test.lua
-- fine ...
luajit -l cunn -e 'nn.testcuda()'
```
)


soumith:
thank you!

PR #250, opened 2016-04-14
Add SpatialClassNLLCriterion

apaszke:
Original `ClassNLLCriterion` is awfully slow when it comes to spatial data. It forces you to transpose it to a 2D Tensor (NCHW -> NHWC), which is pretty expensive by itself for big tensors, and then it uses a **hardcoded number of 32 threads** for processing (e.g. 1M elements)... This means that around 80% of my training loop is spent doing criterion stuff.

Here is a nice comparison from visual profiler:
screen shot 2016-04-14 at 21 52 22
On the left 100x fw+bw pass of SpatialClassNLLCriterion, on the right 1 fw+bw of ClassNLLCriterion (not including transpose).

These are the times for `4x20x512x1024` input and `4x512x1024` targets (100 runs):

| Criterion                                      | Time for 100 runs (s) |
|------------------------------------------------|-----------------------|
| SpatialClassNLLCriterion                       | 0.15400      |
| ClassNLLCriterion                              | 4.03642       |
| Transpose + ClassNLLCriterion + Transpose back | 21.66718       |

If you agree to merge this module here I will also provide Lua code and write a CPU version for `nn`. This can speed up `cudnn.SpatialCrossEntropyCriterion` significantly.

### DIFF
--- a/lib/THCUNN/SpatialClassNLLCriterion.cu
+++ b/lib/THCUNN/SpatialClassNLLCriterion.cu
@@ -1,0 +1,233 @@
+#include "THCUNN.h"
+#include "common.h"
+
+#include <stdio.h>
+#include <assert.h>
+
+// No love for C++11?
+//static_assert(CUDA_NUM_THREADS == CUDA_WARP_SIZE * CUDA_WARP_SIZE,
+//                "SpatialClassNLLCriterion_updateOutput kernel assumes that" \
+//                "number of used threads is a square of warp size");
+
+static const int NWARPS = CUDA_NUM_THREADS / CUDA_WARP_SIZE;
+
+__inline__ __device__ float warpReduceSum(float val) {
+  for (int offset = CUDA_WARP_SIZE/2; offset > 0; offset /= 2)
+    val += __shfl_down(val, offset);
+  return val;
+}
+
+__global__ void cunn_SpatialClassNLLCriterion_updateOutput_kernel(
+          float *output,
+          float *total_weight,
+          float *input,
+          float *target,
+          float *weights,
+          int size_average,
+          int batch_size,
+          int n_classes,
+          int map_nelem,
+          int blocks_per_sample)
+{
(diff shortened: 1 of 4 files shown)


apaszke:
Also, most time for fw kernel is apparently spent in memset calls for [output and total_weight](https://github.com/torch/cunn/pull/250/commits/14657707e2cd4b99d6bfda59b2c2f0372f50146a#diff-5036ca50e84f862d3bd34fc4fa3dc2a4R133), but I don't know how to include this in the main kernel. If anyone has any ideas, just let me know please.

apaszke:
And it would be probably nice to make `ClassNLLCriterion` smarter w.r.t. to block/grid size choices :confused: I can do that later as well.

apaszke:
I adjusted the block assignment and did a small code refractor. The run time has dropped to ~0.13s. There's only one catch - my `updateOutput` kernel assumes that `CUDA_NUM_THREADS == CUDA_WARP_SIZE * CUDA_WARP_SIZE` and won't work otherwise. It would be good to enable some kind of a static assert that checks this, but apparently we're compiling THCUNN with c++11 support disabled. I could use that hacky 'invalid array size' assert, but I don't like it very much. For now I commented it out.

culurciello:
well done

apaszke:
@soumith Any opinions?

soumith:
@apaszke yes opinions. is there test code for correctness checks? If so, can you add it to this PR?

apaszke:
@soumith Yeah, I have a test case that compares this implementation to ClassNLLCriterion, but it also uses a lua part of SpatialClassNLLCriterion (I only don't have a CPU part at the moment). Should I send a full PR to nn first or would you like me to check raw THCUNN functions?

apaszke:
I've sent a PR to `nn` and updated this one to include tests (of course they won't pass for as long as you don't have my `nn` version).

soumith:
Thanks Adam!

PR #260, opened 2016-04-27
adding CUDA version of MultiLabelMarginCriterion + tests

soumith:


soumith:
needs https://github.com/torch/nn/pull/796

PR #269, opened 2016-05-11
Add CMAKE_CXX_FLAGS to rockspec

lukeyeager:
Helps with building on Ubuntu 16.04.

Copies @hughperkins's fix from https://github.com/torch/cutorch/pull/402.


soumith:
Thanks Luke!

PR #272, opened 2016-05-17
Add some doc

hughperkins:


soumith:
awesome! :)

hughperkins:
Thanks! :-)

README.md
```diff
@@ -3,3 +3,95 @@
...
+  * therefore try to allocate any `CudaTensor`s once, at the start of the program,
+  and then simply copy data backwards and forwards
+  between main memory and existing `CudaTensor`s
+* similarly, try to avoid any operations that implicitly allocate new tensors.  For example, if you write:
+```lua
+require 'cutorch'
+
+local a = torch.CudaTensor(1000):uniform()
+for it=1,1000 do
+  local b = a:add(1)
```
fmassa: This is a bad example IMHO, as in here `a:add(1)` returns `a`, meaning that there is no memory allocation (`a` and `b` are the same lua object).
Maybe it would be better something like `b = torch.add(a,1)` or `b = a + 1` ?

hughperkins: Addressed in 7280a028 https://github.com/torch/cunn/pull/273

PR #274, opened 2016-05-18
Limit registers when LRNFillScale runs on TK1.

verdimrc:
Fix for torch@cunn#271, to make openface works on Jetson TK1.


soumith:
Thank you!

Issue #117, opened 2015-07-23
Test failed with th -lcunn -e "nn.testcuda()"


LijieTu:
When I run test.sh, th -lcunn -e "nn.testcuda()"  is unstable and occasionally fails with different error messages:

Every time the error message is different, some examples are: 

SpatialSubSampling_backward
error on state (backward) 
 LT(<) violation   val=1.2421855926514, condition=0.01
	/root/torch/install/share/lua/5.1/torch/Tester.lua:26: in function 'assertlt'
	/root/torch/install/share/lua/5.1/cunn/test.lua:1391: in function 'v'

LogSoftMax_forward_batch
error on state (forward) 
 LT(<) violation   val=0.0010080337524414, condition=0.001
	/root/torch/install/share/lua/5.1/torch/Tester.lua:26: in function 'assertlt'
	/root/torch/install/share/lua/5.1/cunn/test.lua:2364: in function 'v'

I guess a similar issue was raised in https://github.com/torch/cunn/issues/50 and solved(maybe?). 

I updated to the latest torch and packages.

I use a ubuntu 14.04 docker image and cuda 7.0

Thanks.

xonobo:
I have similar test failures. Just updated my torch distro today.

Here is the output of nn.testcuda and cutorch.test outputs:

-------------------------------------------------------------
th> nn.testcuda()
seed: 	1437743792	
Running 94 tests	
==> Done Completed 183 asserts in 94 tests with 2 errors	
SpatialSubSampling_forward_batch
 Function call failed 
...ayci/torch-distro/install/share/lua/5.1/torch/Tensor.lua:243: cuda runtime error (2) : out of memory at /home/bozkalayci/torch-distro/extra/cutorch/lib/THC/THCStorage.cu:30
stack traceback:
	[C]: in function 'resize'
	...ayci/torch-distro/install/share/lua/5.1/torch/Tensor.lua:243: in function 'cuda'
	...kalayci/torch-distro/install/share/lua/5.1/cunn/test.lua:1329: in function 'v'
	...kalayci/torch-distro/install/share/lua/5.1/cunn/test.lua:3496: in function <...kalayci/torch-distro/install/share/lua/5.1/cunn/test.lua:3494>
	[C]: in function 'xpcall'
	...ayci/torch-distro/install/share/lua/5.1/torch/Tester.lua:115: in function 'pcall'
	...ayci/torch-distro/install/share/lua/5.1/torch/Tester.lua:182: in function '_run'
[...]

PR #277, opened 2016-05-20
Fix SpatialSubSampling (was doing non-atomic writes in backprop).

davidsaxton:
Also some changes to test to make it less flaky:
- Decrease some output counts to reduce out-of-memory issues.
- Decrease some tolerances.
- Increase precision of random seed so that multiple test launches at
  the same time start with different seeds.


davidsaxton:
Note partially fixes https://github.com/torch/cunn/issues/117, but SpatialDilatedConvolution is still buggy and fails occasionally (gives all-zero gradInput).

Issue #150, opened 2015-10-01
cunn/SoftMax does not support images with 2^16 or more elements


jvlmdr:
The nn version works
```lua
model = nn.SoftMax()
x = torch.randn(1, 2, 65536, 1)
y = model:forward(x)
```
however the cunn version fails
```lua
model = model:cuda()
x = x:cuda()
y = model:forward(x)
```
with error
```
nngraph/gmodule.lua:281: invalid configuration argument at [...]/torch/extra/cunn/SoftMax.cu:153
```
This error does not occur when the image contains 65535 elements.

hughperkins:
this is related to https://github.com/torch/cunn/issues/144 and https://github.com/torch/cunn/issues/148

dan-sam:
I think this results from using a 2D grid of blocks to call the kernel where the first dimension of the grid is set by the batch size and the second dimension of the grid is set by multiplying the width and height of the "image" (more specifically, the width and height of the input when the SoftMax is computed across the depth).  Since there is a max size of 65535 (2^16 - 1) blocks per dimension in a grid, this caps the number of elements (i.e. width*height) at 65535.

This can be rectified by using a 3D grid of blocks and setting the 2nd/3rd dimension of the grid to the width/height of the image and then adding the 3rd dimension to the computation of the offset from the start of the input/output for each block. I have code that implements this but I think it will only work on GPUs with compute capability of 2.X or greater since the number of dimensions for a grid on a 1.X GPU is capped at 2. Also, I have only really tested that this solution works and have not tested that this solution is efficient.

The CPU spatial SoftMax code does not suffer from this maximum. So another solution is to copy the output from the from right before the Softmax to the CPU and compute the SoftMax on the CPU. Though not ideal, I have found this to be a workable solution in practice since the SoftMax is typically only used once at the end of a network (I think?) and does not make up a major portion of the computation.

andreaskoepf:
'Table 13. Technical Specifications per Compute Capability' of the [Cuda Programming Guide](
http://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#ixzz49EKnp7Yl) says
"Maximum y- or z-dimension of a grid of thread blocks 	65535".

@dmsammons I think your suggestion to use three dimensional blocks is a good idea. 

[...]

PR #279, opened 2016-05-20
Use 3d thread blocks in SoftMax for inputs with w*h > 2^16-1

andreaskoepf:
Fix for https://github.com/torch/cunn/issues/150.
In oder to stay compatible with compute capability 2.X devices a 3D grid is only used for inputs with width*height > 65335.

Todo:
- Analyse performance impact for 3d thread block grid

Would be great if somebody could review/test.

Beside running cunn.test() I compared with the CPU results:

```
require 'cunn'
-- cpu:
s1 = nn.SpatialSoftMax()
x1 = torch.rand(2,16,300,300)

-- gpu:
s2 = nn.SpatialSoftMax()
s2:cuda()
x2 = x1:cuda()

y1 = s1:forward(x1)
y2 = s2:forward(x2)

d = (y1-y2:double()):abs()
err_max = d:max()
err_avg = d:sum()/d:nElement()
print(string.format('err_max: %e, err_avg: %e', err_max, err_avg)) 
```

### DIFF
--- a/lib/THCUNN/SoftMax.cu
+++ b/lib/THCUNN/SoftMax.cu
@@ -5,11 +5,11 @@
 #define SOFTMAX_THREADS 128
 
 __global__ void cunn_SoftMax_updateOutput_kernel(
-  float *output, float *input, int nframe, int dim, int stride)
+  float *output, float *input, int nframe, int dim, int stride0, int stride1)
 {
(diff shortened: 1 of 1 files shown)


soumith:
sorry for the delay, i actually reviewed it because you asked me to :)
It looks good.

PR #285, opened 2016-05-31
Fix test_DataParallelTable.lua

colesbury:
test_DataParallelTable.lua broke because of an incorrect call to assert and changes to the tester and LookupTable.

This fixes the test.


soumith:
Thanks!

PR #294, opened 2016-06-10
Added CUDA version of VolumetricMaxUnpooling

kmul00:
This is the CUDA version of the [this](https://github.com/torch/nn/pull/569) introduced in `nn`.
Added Tests.

modified:   THCUNN.h
new file:   VolumetricMaxUnpooling.cu
modified:   test.lua

### DIFF
--- a/lib/THCUNN/VolumetricMaxUnpooling.cu
+++ b/lib/THCUNN/VolumetricMaxUnpooling.cu
@@ -1,0 +1,246 @@
+#include "THCUNN.h"
+#include "common.h"
+#include "THCDeviceTensor.cuh"
+#include "THCDeviceTensorUtils.cuh"
+#include "THCDeviceUtils.cuh"
+
+#include <cfloat>
+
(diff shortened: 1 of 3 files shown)


lib/THCUNN/VolumetricMaxUnpooling.cu
```diff
@@ -0,0 +1,245 @@
...
+                                             cudaIndices,
+                                             cudaGradInput,
+                                             dT, dH, dW,
+                                             padT, padH, padW, offsetZ);
+    THCudaCheck(cudaGetLastError());
+    totalZ -= 65535;
+    offsetZ += 65535;
+  }
+
+  // cleanup
```
soumith: there's a memory leak. You also need:
THCudaTensor_free(state, input);

kmul00: Yeah, missed it. Thanks !
Done.

soumith:
Thanks!

PR #299, opened 2016-06-16
inplace hardtanh, remove relu6

szagoruyko:
cunn counterpart of https://github.com/torch/nn/pull/851


lib/THCUNN/HardTanh.cu
```diff
@@ -47,13 +69,35 @@ struct hardtanhupdateGradInput_functor
     else
       *gradInput = *gradOutput;
   }
+
+  __device__ void operator()(float *gradInput, const float *input) const
+  {
+    if (*input <= min_val_ || *input >= max_val_)
```
colesbury: Why is this <= and >= but the in-place gradInput operator() call < and > ?

szagoruyko: because in `inplace` mode input is changed and it doesn't have any numbers outside. there can be `==`, I should add a comment about that.

PR #301, opened 2016-06-22
Add noBias for nn.SpatialFullConvolution

PraveerSINGH:
Adds NoBias for nn.SpatialFullConvolution

### DIFF
--- a/lib/THCUNN/SpatialFullConvolution.cu
+++ b/lib/THCUNN/SpatialFullConvolution.cu
@@ -21,6 +21,7 @@ void THNN_CudaSpatialFullConvolution_updateOutput(
 
   THCUNN_assertSameGPU(state, 6, input, output, weight,
                                  bias, columns, ones);
+
   THArgCheck(input->nDimension == 3 || input->nDimension == 4, 2, "3D or 4D (batch mode) tensor is expected");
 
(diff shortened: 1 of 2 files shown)


lib/THCUNN/SpatialFullConvolution.cu
```diff
@@ -19,8 +19,11 @@ void THNN_CudaSpatialFullConvolution_updateOutput(
   int nInputPlane = THCudaTensor_size(state, weight, 0);
   int nOutputPlane = THCudaTensor_size(state, weight, 1);
 
-  THCUNN_assertSameGPU(state, 6, input, output, weight,
-                                 bias, columns, ones);
+  THCUNN_assertSameGPU(state, 5, input, output, weight,
+                                 columns, ones);
+  if (bias) {
```
colesbury: Is this necessary? I thought THCUNN_assertSameGPU handles NULL tensors

PraveerSINGH: Hey @colesbury, You are right, this looks unnecessary. My apprehension was if bias and gradbias, once set to nil, would be converted to NULL tensors or not. However since these are optionalTensors, there seems to be no problem.

PR #305, opened 2016-07-07
Fix BatchNormalization warpSum for pre-Kepler cards

### LINKED ISSUE #298: BatchNormalization.cu fails for large inputs
When using BatchNormalization with large inputs I get an error:
`THCudaCheck FAIL THCUNN/BatchNormalization.cu line=257 error=11 : invalid argument`

I suspect this has a similar cause as https://github.com/torch/cunn/issues/284 when having an input size larger than 65535.

colesbury:
There were two problems with the implementation for pre-Kepler cards:

1. `MAX_BLOCK_SIZE` wasn't accurate which caused crashes if the image size (dim 3 x dim 4) was larger than 256.

2. The `__syncthreads()` in warpSum() caused an incorrect computation (the tests failed) because not every warp in the block executes the `warpSum()` at the end of the [`reduce()`](https://github.com/torch/cunn/blob/86d9f566baa8c0950e41a41f49965ad88611b514/lib/THCUNN/BatchNormalization.cu#L109) function 

Fixes #298

### DIFF
--- a/lib/THCUNN/BatchNormalization.cu
+++ b/lib/THCUNN/BatchNormalization.cu
@@ -7,6 +7,20 @@ const int WARP_SIZE = 32;
 typedef THCDeviceTensor<float, 3> DeviceTensor3;
 typedef THCDeviceTensor<float, 1> DeviceTensor1;
 
+// The maximum number of threads in a block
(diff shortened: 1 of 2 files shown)


soumith:
Thanks!

PR #307, opened 2016-07-17
BCECriterion THCUNN

szagoruyko:
CUDA part of https://github.com/torch/nn/pull/626
It does not support weights right now, two solutions possible:
* have a temp buffer size of input and copy unfolded weights there, can use thrust then
* write a special kernel for BCECriterion

### DIFF
--- a/lib/THCUNN/BCECriterion.cu
+++ b/lib/THCUNN/BCECriterion.cu
@@ -1,0 +1,77 @@
+#include "THCUNN.h"
+#include "common.h"
+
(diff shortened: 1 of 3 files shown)


szagoruyko:
I have to say that the old implementation is terribly slow, just checked that for example dcgan.torch is more than 10% faster with this PR

soumith:
@szagoruyko can you go down the thrust method? right now it is better to not merge this, as there  is not even an assertion on weights. so users will silently think that weights are being considered...

PR #310, opened 2016-07-29
gemm -> Sgemm

soumith:
cutorch introduced the change gemm -> Sgemm and gemv -> Sgemv in master.
Fixing it in cunn.
If one wants to use the old cutorch / cunn, 1.0-0 is a stable previous version without breaking API changes.

```bash
luarocks remove --force cutorch
luarocks remove --force cunn
luarocks install cunn 1.0-0
```


soumith:
Thanks Sergey!

Issue #292, opened 2016-06-07
nn.SpatialSoftMax breaking


shubhamjain0594:
```
require 'nn'
require 'cunn'

softOutCalc = nn.Sequential():add(nn.SpatialSoftMax())
softOutCalc = softOutCalc:cuda()
ips = torch.rand(8,2,80,80)
ips = ips:cuda()
softOutCalc:forward(ips)
```

This code gives the error 

> THCudaCheck FAIL file=/tmp/user/1012/luarocks_cutorch-scm-1-199/cutorch/lib/THC/generic/THCTensorCopy.c line=81 error=77 : an illegal memory access was encountered
/home/users/shubhamj/torch/install/bin/luajit: cuda runtime error (77) : an illegal memory access was encountered at /tmp/user/1012/luarocks_cutorch-scm-1-199/cutorch/lib/THC/generic/THCStorage.c:158

It used to work fine on earlier installations.

fmassa:
I just tried the snippet using most recent installations of `cutorch` and `cunn` and it works fine for me.

shubhamjain0594:
I have updated to the latest version. Still getting same errors. Any help is appreciated. 
I am working with cuda7.5 and latest installs of torch, cutorch, nn, cunn and cudnn.

soumith:
Shubham, is this still a blocker? if so i can have a look.

Give me info on your GPU model (of all GPUs on system) and NVIDIA driver.

shubhamjain0594:
Its working fine. Thanks

chsasank:
I have the exact same issue. So, reopening the issue.

chsasank:
Let me give more cases with cudnn:

```
require 'nn'
require 'cunn'
require 'cudnn'

softOutCalc = nn.Sequential():add(nn.SpatialSoftMax())
softOutCalc = softOutCalc:cuda()
softOutCalc = cudnn.convert(softOutCalc, cudnn)  
ips = torch.rand(8,2,80,80)
ips = ips:cuda()
softOutCalc:forward(ips)
```
Gives
```
THCudaCheck FAIL file=/home/users/sasank/torch/extra/cutorch/lib/THC/generic/THCStorage.c line=147 error=77 : an illegal memory access was encountered
```

While, 
```
require 'nn'
require 'cunn'
require 'cudnn'

softOutCalc = nn.Sequential():add(cudnn.SpatialSoftMax())
softOutCalc = softOutCalc:cuda()
[...]

PR #312, opened 2016-08-02
fix SpatialSoftMax bug and add unit tests

soumith:


soumith:
cc: @andreaskoepf

soumith:
Fixes: https://github.com/torch/cunn/issues/292

PR #314, opened 2016-08-05
Fix "invalid configuration" when using very large batch sizes in evaluate mode

colesbury:
Example:
```
 bn = nn.BatchNormalization(100):cuda()
 bn:evaluate()
 bn:forward(torch.CudaTensor(147000, 100):zero())
 cutorch.synchronize()
```

Fixes https://github.com/torch/nn/issues/907

Per-activation batch norm still needs to be optimized. It currently runs with one thread per block. Improving per-activation performance is left for a future PR.


soumith:
Thanks Sam!

PR #317, opened 2016-08-11
CUDA version of Spatial Dilated Max Pooling

kmul00:
`CUDA` version of [this](https://github.com/torch/nn/pull/921).

modified:   lib/THCUNN/SpatialMaxPooling.cu
modified:   lib/THCUNN/THCUNN.h
modified:   test.lua


soumith:
rebased and merged via https://github.com/torch/cunn/commit/25498b73c00ed3a25d29f476f1ea0df81e23d7d6

PR #318, opened 2016-08-11
Improvements/fixes in THCUNN

apaszke:


soumith:
Thanks!

lib/THCUNN/DistKLDivCriterion.cu
```diff
@@ -64,7 +64,7 @@ void THNN_CudaDistKLDivCriterion_updateGradInput(THCState *state, THCudaTensor *
              "input and target need to have the same number of elements");
 
   long size = THCudaTensor_nElement(state, input);
-  float norm = (sizeAverage ? 2./size : 2.);
+  float norm = (sizeAverage ? 1./size : 1.);
```
albanD: So the gradients for this criterion has been off by a factor of two since its creation? [initial commit](https://github.com/torch/cunn/commit/95f7a5734fa474518f070f8ba90ab16506a9e5db#diff-d6a9bf748d07d407e9a3946fb1c8c1d7R68)

soumith: yep :-/
only reason we never caught it was because cunn backward tolerance has been 1e-2 since the beginning of time.

PR #326, opened 2016-08-25
Consistent Max Pool API

kmul00:
`CUNN` version of [this](https://github.com/torch/nn/pull/941).

renamed:    lib/THCUNN/SpatialMaxPooling.cu -> lib/THCUNN/SpatialDilatedMaxPooling.cu
modified:   lib/THCUNN/SpatialMaxPooling.cu
modified:   lib/THCUNN/THCUNN.h


soumith:
Thanks a lot!

PR #329, opened 2016-09-01
inplace is reversed in HardTanh:backward.

gchanan:
Fixes torch7 issue #734, "Inconsistent behavior of nn.Clamp in CPU and GPU modes".
Adds a simple test that gradOutput equals gradInput after backward when inplace is set.
It is possible to construct a test with inplace HardTanh where forward+backward yields
different results for nn vs cunn, but this appears to be due to the inclusive vs
exclusive bounds used for inplace vs non-inplace, respectively, so a more direct test
is preferred.


soumith:
Nice catch! Thanks

PR #331, opened 2016-09-01
BCECriterion THCUNN + Weights

nitsky:
Adds weights support to BCECriterion THCUNN implementation.

Rather than writing a special kernel, I followed [this thrust example](https://github.com/thrust/thrust/blob/master/examples/arbitrary_transformation.cu) to perform a transformation of more than two inputs, replacing `thrust::inner_product` with `thrust::transform_reduce`.

I added two new tests so BCECriterion is tested both with and without weights like in torch/nn.

### DIFF
--- a/lib/THCUNN/BCECriterion.cu
+++ b/lib/THCUNN/BCECriterion.cu
@@ -1,40 +1,55 @@
 #include "THCUNN.h"
 #include "common.h"
 
-#include <thrust/fill.h>
 #include <thrust/functional.h>
 #include <thrust/device_ptr.h>
-#include <thrust/reduce.h>
-#include <thrust/inner_product.h>
+#include <thrust/iterator/zip_iterator.h>
+#include <thrust/transform.h>
+#include <thrust/transform_reduce.h>
 
 const float eps = 1e-12f;
 
 struct bce_functor
 {
-  __host__ __device__ float operator()(const float& x, const float& y) const
+  template <class Tuple>
+  __host__ __device__
+  float operator()(Tuple x)
   {
-    return - logf(x + eps) * y - logf(1.f - x + eps) * (1.f - y);
+    float o = thrust::get<0>(x);
+    float t = thrust::get<1>(x);
+    float w = thrust::get<2>(x);
+    return - w * (t * logf(o + eps) + (1.f - t) * logf(1.f - o + eps));
   }
 };
 
 void THNN_CudaBCECriterion_updateOutput(THCState *state, THCudaTensor *input, THCudaTensor *target, THCudaTensor *output, bool sizeAverage, THCudaTensor *weights)
 {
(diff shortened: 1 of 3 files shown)


szagoruyko:
BCECriterion accepts weights which have different shape than input and there is an expandAs call here https://github.com/torch/nn/pull/626/files#diff-644e6014799c3b6484776f24649e693bR30, so this line here will be allocating a new tensor on each forward call https://github.com/torch/cunn/pull/331/files#diff-e667e2a92577d901f42c8ede2fa009fdR33

besides, you need to keep the non-weighted path as weight tensor is optional and might be NULL

nitsky:
@szagoruyko I pushed changes to restore the un-weighted code path, but I'm not sure how to prevent allocating a new tensor on each forward call, do you know?

szagoruyko:
@nitsky see my original comment in https://github.com/torch/cunn/pull/307, either a special kernel, or we need to cache the temporary contiguous tensor in lua side

lib/THCUNN/BCECriterion.cu
```diff
@@ -0,0 +1,162 @@
...
+    float t = thrust::get<1>(x);
+    float w = thrust::get<2>(x);
+    return - (t - o) / ((1 - o + eps) * (o + eps)) * norm * w;
+  }
+};
+
+void THNN_CudaBCECriterion_updateGradInput(THCState *state, THCudaTensor *input, THCudaTensor *target, THCudaTensor *gradInput, bool sizeAverage, THCudaTensor *weights)
+{
+  if (weights) {
+    THCUNN_assertSameGPU(state, 4, input, target, gradInput, weights);
```
szagoruyko: THCUNN_assertSameGPU can handle NULL, no need for if here

szagoruyko:
@nitsky I'll add the caching with CudaTensor condition in https://github.com/torch/nn/pull/626, and this will be good to go.

soumith:
@szagoruyko ok that sounds good. I am not sure why the weights are non-contiguous...
Either ways, this is a criterion, so it seems like alloc / free seems okay, especially at a time when we are going to push the new memory allocator soon

nitsky:
@soumith I have added the test against the BCECriterion lua implementation. @szagoruyko I removed the unnecessary weights check for assertSameGPU.

test.lua
```diff
@@ -3053,6 +3053,128 @@ end
 
 -- Criterion tests
 
+function BCECriterion_forward_truth(buffer, input, target, weights, sizeAverage)
```
soumith: local function ....

lua leaks variables as global by default :*(

nitsky:
oops :) done.

szagoruyko:
@soumith let's merge it like that then, thanks @nitsky !

soumith:
Thanks @nitsky

PR #332, opened 2016-09-06
Fix issue with flatten parameters in DataParallelTable

vivekn:
Currently, if only a part of a network is wrapped with DataParallelTable and flattenParameters is enabled, then the flat tensor created is larger than it needs to be due to an issue with offsets. This ends up copying a lot more data during every call to syncParameters.


soumith:
Thank you!

PR #338, opened 2016-09-26
SpatialLogSoftMax

nitsky:
I updated LogSoftMax to support spatial data with a kernel that accesses the input data in "cuda" format, B x H x W x P, where B is batch size and P is number of classes, for better cache performance. The original commit from @fmassa is included in this PR.

### DIFF
--- a/lib/THCUNN/LogSoftMax.cu
+++ b/lib/THCUNN/LogSoftMax.cu
@@ -1,104 +1,80 @@
 #include "THCUNN.h"
 #include "common.h"
 
-#define LOGSOFTMAX_THREADS 128
-
-__global__ void cunn_SpatialLogSoftMax_updateOutput_kernel(float *output, float *input, int nframe, int dim, int stride)
+__global__ void cunn_SpatialLogSoftMax_updateOutput_kernel(float *output, float *input, int classSize, int height, int width)
 {
-  __shared__ float buffer[LOGSOFTMAX_THREADS+1];
-  int k = blockIdx.x;
-  float *input_k = input + k*dim*stride + blockIdx.y;
-  float *output_k = output + k*dim*stride + blockIdx.y;
-  int tx = threadIdx.x;
+  int batchIndex = blockIdx.x;
+  int index = threadIdx.x;
 
-  int i_start = threadIdx.x;
-  int i_end = dim;
-  int i_step = blockDim.x;
+  while (index < height*width) {
+    int y = index / width;
+    int x = index % width;
+    if (y >= height)
+      break;
 
-  // max?
-  buffer[tx] = -FLT_MAX;
-  for (int i=i_start; i<i_end; i+=i_step)
-  {
-    float z = input_k[i*stride];
-    if(buffer[tx] < z)
-      buffer[tx] = z;
+    // calculate input starting index in cuda layout (B x H x W x P)
+    int inputStartIndex = 0
+    + (height*width*classSize)*batchIndex
+    + (width*classSize)*y
+    + (classSize)*x;
+
+    // compute a
+    float a = 0;
+    for (int i = 0; i < classSize; i++) {
+      a += __expf(input[inputStartIndex + i]);
+    }
+
+    // invert a
+    a = 1.0f / a;
+
+    for (int i = 0; i < classSize; i++) {
+      // calculate output index in torch layout (B x P x H x W)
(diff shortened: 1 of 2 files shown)


lib/THCUNN/LogSoftMax.cu
```diff
@@ -1,6 +1,81 @@
...
+    + (classSize)*x;
+
+    // compute a
+    float a = 0;
+    for (int i = 0; i < classSize; i++) {
+      a += __expf(input[inputStartIndex + i]);
+    }
+
+    // invert a
+    a = 1.0f / a;
```
apaszke: This really doesn't need this kind of comment (same goes for `// compute a`) 😉 If you want to add some insights, it would be more helpful to say what a is (e.g. show it in a softmax formula). It's not a very good name though, it's not conventionally used anywhere in context of Softmax.

lib/THCUNN/LogSoftMax.cu
```diff
@@ -1,6 +1,81 @@
...
+
+  while (index < height*width) {
+    int y = index / width;
+    int x = index % width;
+    if (y >= height)
+      break;
+
+    // calculate output starting index in cuda layout (B x H x W x P)
+    int outputStartIndex = 0
+    + (height*width*classSize)*batchIndex
```
apaszke: nit: indent the additions (in multiple places in this file)

lib/THCUNN/LogSoftMax.cu
```diff
@@ -1,6 +1,81 @@
...
+    + (width*classSize)*y
+    + (classSize)*x;
+
+    // compute sum
+    float a = 0;
```
apaszke: if it's a sum, maybe it's better to call it `sum`

soumith:
The PR has lots of memory leaks via refcounting mistakes.
Whenever you call a TH*Tensor_new*(...), you have to remember to TH*Tensor_free(...) the same tensor...
Can you fix those please.

lib/THCUNN/LogSoftMax.cu
```diff
@@ -1,6 +1,81 @@
...
+    for (int i = 0; i < classSize; i++) {
+      // calculate output index in torch layout (B x P x H x W)
+      int outputIndex = 0
+      + (classSize*height*width)*batchIndex
+      + (height*width)*i
+      + (width)*y
+      + x;
+      output[outputIndex] = logf(a * __expf(input[inputStartIndex + i]));
+    }
+    index += 1024;
```
apaszke: Use `blockDim.x` in here

lib/THCUNN/LogSoftMax.cu
```diff
@@ -1,6 +1,81 @@
...
+  int index = threadIdx.x;
+
+  while (index < height*width) {
+    int y = index / width;
+    int x = index % width;
+    if (y >= height)
+      break;
+
+    // calculate output starting index in cuda layout (B x H x W x P)
+    int outputStartIndex = 0
```
apaszke: nit: this 0 serves no purpose, just start with the first term.

lib/THCUNN/LogSoftMax.cu
```diff
@@ -256,36 +331,100 @@ void THNN_CudaLogSoftMax_updateOutput(THCState *state, THCudaTensor *input, THCu
...
+    THCudaTensor *transposedInput = THCudaTensor_new(state);
+    THCudaTensor_resizeAs(state, transposedInput, input);
+    THCudaTensor_copy(state, transposedInput, input);
+    input = transposedInput;
+    // convert tensor from torch layout to cuda layout
+    // P x H x W -> W x H x P
+    THCudaTensor_transpose(state, input, input, 0, 2);
+    // W x H x P -> H x W x P
+    THCudaTensor_transpose(state, input, input, 0, 1);
+    input = THCudaTensor_newContiguous(state, input);
```
apaszke: Rather than doing a clone of input + transpose + contiguous, which executes two copies, it would be better to create `transposedInput` already with correct sizes and copy `input` into it (copy only cares about the number of elements, not about the dimension sizes). Same goes for transpose in `updateGradInput`.

nitsky:
@soumith @apaszke I believe I have resolved the issues you brought up in the latest commit. Thanks for your help.

soumith:
Thanks David!

PR #344, opened 2016-10-03
Generic soft plus

gchanan:
This supports generic SoftPlus.

There is some additional code/complexity in supporting both generic and non-generic modules simultaneously; it probably makes the most sense to convert everything to generic before doing the cleanup, but let me know if you feel differently.  These are:
- THCUNN.lua/CMakeLists needs to know about both the generic and non-generic libraries
- THCAutoHalfNumerics could move to THC (we probably will need a lot of additional operators/conversions here, so I'd rather modify it in one repo for now)
- there are multiple versions of assertGPU


soumith:
#346 is a superset

PR #346, opened 2016-10-07
Add generic support for pointwise modules

gchanan:
Adds generic (half/float/double) support for pointwise modules.

I made notes about questions/judgement calls in the commit messages for the specific modules that may be helpful for reviewing purposes.  For example, see the message about __expf in SoftMax.


soumith:
this needs a rebase because i merged in https://github.com/torch/cunn/pull/338 i think

gchanan:
The title is now out of date; this pull request now includes code for all cunn modules, not just pointwise ones.  A couple of other notes:
1) This removes support for non-generic modules.  All new modules must be generic.
2) This changes the tests (at least the ones that test C modules) to run over half, float, double types.  So the test runs may be approximately 3 times as long.

PR #350, opened 2016-10-13
change to work on windows && replace long with ptrdiff_t

BTNC:
Hi,

This PR is to get cunn compile on windows with msvc. It also contains the ptrdiff_t replacement change.
Most tests passed with several random failures. It is like wining a lottery to see PReLU_backward test pass, its error is always "error on weight LT failed: nan >= 0.01". I found nan in tensors during the test when I print the tensors. Do you see similar test failures on *nix systems or can you suggest what maybe the cause?

Thanks,


soumith:
Thanks!
cc: @gchanan you might have to rebase #346 past this

PR #354, opened 2016-10-18
change to work with msvc && ptrdiff_t replacement

BTNC:
Hi,

Most tests could pass with a few random failures. PReLU_backward consistenly fail with following error:
error on weight
LT failed: nan >= 0.01

Thanks,


BTNC:
Update: it seems codes have been fixed that all tests can pass on windows.

BTNC:
Hi, @soumith , cunn is being extensively refactored. When will the refactor work finish, so that I can rework the windows patch?

soumith:
@BTNC i had indeed not merged this in because of the BIG refactor that was ongoing.
Now it is finished, I have merged all the bits. You can rework the windows patch now.

PR #364, opened 2016-11-02
test on inputGpu emptiness symmetric in backward and forward

jsenellart-systran:
I had a problem in `updateGradInput` method for small batches for which one of the gpu assignment was empty. 
In `updateOutput` method, emptiness test is done using `_hasData` which includes empty tensor testing but also **empty table testing** (my net is taking table input).

The patch just does the same for `__backward` - let me know if you have any question.


soumith:
Thanks!

PR #365, opened 2016-11-03
more robust support of small batches when input/output dimensions differ

jsenellart-systran:
I found another problematic context: when input & output dimension differs (for instance input is table, and output is tensor) - the hack of sending back inputGpu[i] does not work and we hit problems in _concat.

I propose the attached patch - which is simpler and avoid sending back inputGpu[i] as output for empty batch assignments.

_distribute returns the maximum gpuIdx index necessary and _exec have an optional parameter to launch process up to the gpuIdx index.

note that this type of issue shows while introducing data parallelism on NMT engine - the batches have identical length sentences, so it happens quite often that we have batches with very few items.

### DIFF
--- a/DataParallelTable.lua
+++ b/DataParallelTable.lua
@@ -188,17 +188,13 @@ function DataParallelTable:updateOutput(input)
    local prevGpuid = cutorch.getDevice()
 
    -- distribute the input to GPUs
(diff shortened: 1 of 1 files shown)


soumith:
thanks!

PR #368, opened 2016-11-10
Protect half operations with CUDA_HALF_TENSOR with generic modules.

gchanan:
This should address https://github.com/torch/torch7/issues/832.

I tested compilation with CUDA 7.0; it fails without and passes with this PR.


soumith:
Thanks

PR #373, opened 2016-11-11
more improvments on error messages and shape checks.

gchanan:


lib/THCUNN/generic/SpatialConvolutionLocal.cu
```diff
@@ -19,6 +85,14 @@ void THNN_(SpatialConvolutionLocal_updateOutput)(
...
                        bias, finput);
 
+  int freeWeight = THNN_(view_weight_local)(state, &weight);
+
+  THNN_(SpatialConvolutionLocal_shapeCheck)
+       (state, input, NULL, weight, bias, kH, kW, dH, dW, padH, padW,
+        inputHeight, inputWidth, outputHeight, outputWidth);
+
+  input = THCTensor_(newContiguous)(state, input);
+
```
soumith: remove the comment here

PR #377, opened 2016-11-15
Add contiguous checks / auto contiguous

gchanan:
These changes were required to get cunn tests working when using all contiguous tensors (except in cases where an arg check complains that a tensor is not contiguous).

The test I used is here (and is not included in this pull request):
https://github.com/gchanan/cunn/commit/ae7a8363c683689087d8457e67fb50aebb923469

I'm happy to incorporate the test into a commit, but it's not obvious to me how to do that in a clean way (i.e. that doesn't complicate all the existing tests).


gchanan:
All input/gradOutput tensors are now made non contiguous (have holes) and modules call newContiguous on tensors that are required to be contiguous.

PR #381, opened 2016-11-26
Fix shapeCheck in Spatial Pooling modules

fmassa:
Cuda version of https://github.com/torch/nn/pull/1049


soumith:
awesome! thanks.

PR #389, opened 2016-12-01
Improve fix/shape checks for Temporal and Spatial modules

gchanan:
1. Adds gradOutput shape checks for temporal modules
2. Removes unnecessary (too restrictive and/or redundant) shape checks in Spatial Pooling modules.

### DIFF
--- a/lib/THCUNN/generic/SpatialAveragePooling.cu
+++ b/lib/THCUNN/generic/SpatialAveragePooling.cu
@@ -27,11 +27,6 @@ static inline void THNN_(SpatialAveragePooling_shapeCheck)(
 
   THCUNN_argCheck(state, ndim == 3 || ndim == 4, 2, input,
                   "3D or 4D input tensor expected but got: %s");
-
-  THArgCheck(input->size[dimw] >= kW - padW && input->size[dimh] >= kH - padH, 2,
-             "input image (H: %d, W: %d) smaller than kernel "
-             "size - padding( kH: %d padH: %d kW: %d padW: %d",
-             input->size[dimh], input->size[dimw], kH, padH, kW, padW);
   THArgCheck(kW/2 >= padW && kH/2 >= padH, 2,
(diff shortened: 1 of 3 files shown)


lib/THCUNN/generic/SpatialAveragePooling.cu
```diff
@@ -27,11 +27,6 @@ static inline void THNN_(SpatialAveragePooling_shapeCheck)(
 
   THCUNN_argCheck(state, ndim == 3 || ndim == 4, 2, input,
                   "3D or 4D input tensor expected but got: %s");
-
```
gchanan: Let's take the most general case involving padding and dilation.  I believe the intention of the check is to make sure that the effective input size (inputsize + 2 * padding) is >= the effective kernel size (dilation * (k-1) + 1).  If that's the case, the removed check (inputsize >= k - pad) is wrong, because e.g. inputsize=1, pad=1, k=3, dilation=1 is a valid input but fails the check (and runs successfully if the check is removed).

Then, the "corrected" check above is equivalent to the existing check below that the output dimension is >= 1.

gchanan: do you agree with this argument, @soumith?

fmassa: I think this PR is good.
There was a copy/paste problem in https://github.com/torch/cunn/commit/b923c23d4c058e3c4b4784a6c67752967ab8bbf3#diff-bd0935145314cc2aeb3c96b97076eb80L40 that introduced the shapeCheck problem.

BTW, there is still a reminiscent extra shapeCheck in the C counterpart https://github.com/torch/nn/blob/master/lib/THNN/generic/SpatialAveragePooling.c#L136-L139

soumith: yes, i agree :)

Issue #286, opened 2016-05-31
test_DataParallelTable Failing


stephenrawls:
Hi,

I'm running into a problem using more than 2 GPU's when using DataParallelTable. I think my underlying problem probably relates to my failing DataParallelTable tests, copied below:

```
% th <snip>/install/share/lua/5.1/cunn/test_DataParallelTable.lua
Running 13 tests
 1/13 DataParallelTable_emptyTensor ..................................... [ERROR]
 2/13 DataParallelTable_noGradInput ..................................... [ERROR]
 3/13 DataParallelTable_misc ............................................ [PASS]
 4/13 DataParallelTable_serialize ....................................... [ERROR]
 5/13 DataParallelTable_type ............................................ [PASS]
 6/13 ProfileDataParallelTable .......................................... [WAIT]
Profiling DataParallelTable with 1 gpus   
  Time per FPROP+BPROP: 0.36483557224274  
Profiling DataParallelTable with 2 gpus   
  Time per FPROP+BPROP: 0.6190810918808   
Profiling DataParallelTable with 3 gpus   
  Time per FPROP+BPROP: 0.53080530166626  
Profiling DataParallelTable with 4 gpus   
  Time per FPROP+BPROP: 0.5239892244339   
Profiling DataParallelTable with 5 gpus   
 6/13 ProfileDataParallelTable .......................................... [ERROR]
 7/13 DataParallelTable_sync ............................................ [ERROR]
 8/13 DataParallelTable_emptyData ....................................... [FAIL]
 9/13 DataParallelTable_streams ......................................... [WAIT]
```

Note that I have to kill the test process because it freezes on test #9. This is the same behavior that I have when training on more than 2 GPU's...the training process eventually freezes, although all GPU's are continually being utilized 100%.

Not sure if this helps, but note that I recently re-built torch from scratch, and still get this error.

Any assistance would be appreciated. I'm available to assist in debugging with a little guidance.

Thanks,
Stephen

Chetter2:
@stephenrawls , did you solve it?

stephenrawls:
@Chetter2 Yes, it ended up being a hardware problem. After working with our hardware vendor, they advised us to update the firmware and change some bios setting or other, after which things worked fine.

PR #390, opened 2016-12-02
Fix size mismatch when using DataParallelTable

iaalm:
When using DataParallelTable on more then 2 GPU, an 'size mismatch' error occurred.
This error generated when the dim_size of input tensor is small than number of GPUs devided by 2.
When using 4 GPU, test_DataParallelTable.lua also raise two error.
Issue #286 may caused by this bug too.
I think this patch fix that.


cliffwoolley:
Does https://github.com/torch/cunn/pull/397 also fix that issue for you?

iaalm:
Yes, it also works.
I think this PR can be closed.

PR #395, opened 2016-12-08
Add cudaHalf, cudaDouble functions to nn.Module.

gchanan:
This is inconsistent with cuda(), which is included from nn not cunn, but this seems more appropriate since we can check cutorch.hasHalf before adding cudaHalf() (which is consistent with how tensors conversion functions are handled).

### DIFF
--- a/test.lua
+++ b/test.lua
@@ -5439,6 +5439,25 @@ function cunntest.VolumetricReplicationPadding_backward()
    end
 end
 
+function cunntest.ModuleConversionFunctions()
+   local module = nn.Tanh() -- arbitrary module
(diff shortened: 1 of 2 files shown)


test.lua
```diff
@@ -5439,6 +5439,25 @@ function cunntest.VolumetricReplicationPadding_backward()
    end
 end
 
+function cunntest.ModuleConversionFunctions()
+   local module = nn.Tanh() -- arbitrary module
+   input = torch.randn(10)
```
soumith: local input

gchanan:
thanks for catching that, fixed.

PR #399, opened 2016-12-13
Check that batch size matches the target size in ClassNLLCriterion

apaszke:


soumith:
thanks!

JoostvDoorn:
This pull request breaks the single target CrossEntropy in cuda.

See:
```lua
require 'cunn'
criterion = nn.CrossEntropyCriterion():cuda()
criterion:forward(torch.Tensor(3, 40):cuda(), torch.Tensor{1,1,1}:cuda())
criterion:forward(torch.Tensor(1, 40):cuda(), torch.Tensor{1}:cuda())
```

The test ClassNLLCriterionSingleTarget was already broken before this pull request.

soumith:
@JoostvDoorn thanks. this surfaced a bug in nn, i fixed it via https://github.com/torch/nn/commit/3b8e5a8064b4d359ab691369ace524c9b5b87575

JoostvDoorn:
Great thanks!

PR #400, opened 2016-12-15
Add missing free in LookupTable

apaszke:
There are two exit points in `accGradParameters` and one of them is missing a `free` on gradOutput (`newContiguous` is called on it [at the beginning](https://github.com/torch/cunn/compare/master...apaszke:lookup_leak?expand=1#diff-5d6c9b1c7b1eb8b369bb74867cb73867R18)).


soumith:
Thanks

PR #401, opened 2016-12-16
Add support for Double tensors to DataParallelTable.

gchanan:
Also adds currently disabled support for Half tensors,
which requires torch.HalfTensor to enable.


soumith:
thanks!

PR #414, opened 2017-01-10
Re-route thrust memory allocation to THCudaMalloc / THCudaFree in cunn.

gchanan:


gchanan:
Note this requires https://github.com/torch/cutorch/pull/668.

PR #418, opened 2017-01-15
Add SpatialAdaptiveAveragePooling.

ruotianluo:
The code is based on SpatialAdaptiveMaxPooling. This use the same pool region algorithm as SpatialAdaptiveMaxPooling.


ruotianluo:
torch/nn#1102

soumith:
thanks!

PR #419, opened 2017-01-17
Fixing HalfTensor FIXME

borisfom:
Now that we have torch.HalfTensor ...


borisfom:
This is ready for merge.

borisfom:
@soumith: Why aren't we mergiung this ?

PR #423, opened 2017-01-23
Add contiguity checks to THCUNN

apaszke:


soumith:
does this hold for THNN too?

apaszke:
No, I think THNN calls `THTensor_(gemm)`, and it should do all the checks. THCUNN had these bugs, because it directly calls into cuBLAS, bypassing the tensor interface.

PR #433, opened 2017-01-30
Fix memory leak in SpatialMaxUnpooling

apaszke:


lib/THCUNN/generic/SpatialMaxUnpooling.cu
```diff
@@ -46,7 +46,7 @@ void THNN_(SpatialMaxUnpooling_updateOutput)(
     THCTensor_(resize3d)(state, output, nInputPlane, oheight, owidth);
 
   THCTensor_(free)(state, input);
-
+  THCudaLongTensor_free(state, indices);
```
fmassa: nit: maybe use `THCIndexTensor`?

PR #437, opened 2017-02-02
Revert "cuda implementation of Gated Linear Unit, fixed issues with genericization"

soumith:
Reverts torch/cunn#430

@huihuifan second time you broke trunk with the same PR :)
Please test the PRs locally before you send them out with: `luarocks make rocks/cunn-scm-1.rockspec`


gchanan:
You need to include THCApply like in this commit: https://github.com/torch/cunn/commit/22b109d8a0ef5e1627dbae87bbd7790940766d02

soumith:
@huihuifan it is not your fault, it looks like in between when you tested the PR and when I pushed it in, the ground was moved from under.

PR #438, opened 2017-02-03
cuda implementation of Gated Linear Unit, fixed issues with genericiz…

huihuifan:
…ation (ran `luarocks make rocks/cunn-scm-1.rockspec` which built completely)
fixed error in #430


soumith:
thanks angela!

PR #446, opened 2017-02-21
[DPT] Add option for synchronous copy of model to GPUs

lvdmaaten:
There exist use-cases in which the serialization code for a model is not thread-safe, for instance, when the serialization code has non-thread-safe dependencies.

This PR adds an option to perform synchronous copies of the model onto all the GPUs in order to prevent race conditions. This change should not break existing code.


soumith:
thanks!

PR #452, opened 2017-03-20
Add support of depth-wise convolution

stooloveu:
See: https://github.com/torch/nn/pull/1098


nicholas-leonard:
@stooloveu Added unit test, fixed a bug, rebased and merged https://github.com/torch/cunn/commit/e97095d33184f97b5c82e606031071fd107b298e. Thank you for your patience and hard work!

PR #454, opened 2017-03-24
Fix inconsistent in-place and out-of-place for HardTanh

ngimel:
in-place and out-of-place updateGradOutput results are different where input=min_val or input=max_val


soumith:
thanks!

PR #455, opened 2017-03-24
Adding Indexlinear

pavanky:
The cunn implementation of IndexLinear described here: https://github.com/torch/nn/pull/1181


soumith:
thanks!

PR #458, opened 2017-04-05
Update to ignore zero targets

jnhwkim:
If the target is zero, loss and gradient of input are reminded to be zero. It
is useful for variable-length NL generation models.


nicholas-leonard:
@jnhwkim Thanks!

jnhwkim:
I think `nn` should be consistent with this. I'll look into it, shortly.

nicholas-leonard:
@jnhwkim I am working on nn version of this: https://github.com/torch/nn/pull/1215

PR #459, opened 2017-04-07
Support TORCH_NVCC_FLAGS environment variable

triccardi-systran:
This is already supported in cutorch since august 2016, and is used in
pytorch integration (to reduce the binary size: pytorch/pytorch#436).

Cherry-pick torch/cutorch@1c2e80c5e5247a30899f7113a1dac77872aaf584


soumith:
thanks!

soumith:
and nice find. I guess pytorch's next bin release will be smaller

triccardi-systran:
With this and `TORCH_NVCC_FLAGS="-Xfatbin -compress-all"` `libTHCUNN.so` size is reduced from 191MB to 37MB.

soumith:
that's awesome! cant believe I missed this. thanks for this!

PR #470, opened 2017-05-24
Fix the mix-up of height and width on depth-wise convolution

qqning:


soumith:
at this rate, is this also a bug?: https://github.com/torch/nn/blob/master/lib/THNN/generic/SpatialDepthWiseConvolution.c#L199

qqning:
Yes! Should I update this pr ?

soumith:
thanks!

PR #477, opened 2017-07-07
GPU implementation of L_p feature pooling

wickedfoo:
This diff contains the GPU kernels and calling code from the fbcunn library for FeatureLPPooling that we are contributing to cunn itself, converted to compile with modern cutorch / cunn with half, float and double support.

https://github.com/facebook/fbcunn/blob/master/src/FeatureLPPooling.cu
https://github.com/facebook/fbcunn/blob/master/src/FeatureLPPoolingHost.cpp

@soumith said he'd add the rest from here.

### DIFF
--- a/lib/THCUNN/FeatureLPPooling.cu
+++ b/lib/THCUNN/FeatureLPPooling.cu
@@ -1,0 +1,653 @@
+#include "THCUNN.h"
+#include "THCAtomics.cuh"
+#include "THCDeviceTensor.cuh"
+#include "THCDeviceTensorUtils.cuh"
+#include "THCDeviceUtils.cuh"
+#include "THCNumerics.cuh"
+#include "THCTensorTypeUtils.cuh"
+
+#define OUTPUT_FEATURES_PER_THREAD 32
+#define MAX_WARPS_PER_RUN 4
+
+namespace detail {
+
+/// Various utilities for dealing with arrays of values which are
+/// maintained in thread-local registers. All accesses are done in such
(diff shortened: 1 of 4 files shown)


wickedfoo:
Missing some updates, one min

wickedfoo:
ok updated with the header too, and float -> accreal

wickedfoo:
Please advise if there's anything else that needs to be added on the cunn side. There was no CPU implementation, only this in fbcunn for the module:

https://github.com/facebook/fbcunn/blob/master/fbcunn/FeatureLPPooling.lua

wickedfoo:
I'm going to write a simple CPU version of this for torch/nn, that shouldn't be too hard.

soumith:
i think we can make do with the Lua version, dont waste your time.

wickedfoo:
There's no implementation of the algorithm in Lua either, just the module file that forwards to the cunn version for CudaTensors. Is that good enough?

soumith:
oh i see. yea it needs a cpu or just pure lua script implementation

wickedfoo:
I added a CPU implementation in https://github.com/torch/nn/pull/1259

I'll update this cunn pull request with a test of the GPU code that compares against the CPU implementation

wickedfoo:
This diff has been updated with a test, bugs fixed and it works, compared against the CPU version.
This is good to go, as is https://github.com/torch/nn/pull/1259.

wickedfoo:
ping... can this be merged?

PR #479, opened 2017-07-24
Added cunn tests for UpSampling module.

mikepound:
Cunn tests for the recently added UpSampling module, have tested and they build / succeed.

Mike


soumith:
thanks mike!

PR #489, opened 2017-11-21
Try to apply softmax to a batch of data with variable length

hfxunlp:
Hi, I want to make SoftMax support variable length input, so you can use a batch of data with different length as the input of this module. This is helpful for Natural Language Processing, especially for the Attention model of seq2seq and Attention-over-Attention model for reading comprehension. And this pull request is corresponding to https://github.com/torch/nn/pull/1297.


hfxunlp:
Sorry, There is still a bug that I have to fix

hfxunlp:
I fix the bug in a weird way in the second commit. It will be wonderful if someone can tell me why and show a more beautiful solution.

